PR #33, opened 2015-12-25
minor document fix

palamccc:


matteocontrini:
I think the argument name was wrong, not the line you modified

matteocontrini:
No sorry,

palamccc:
just tested it, its getting only jobId as argument, not job.

matteocontrini:
Yes sorry, you're right

LewisJEllis:
Thanks!

PR #53, opened 2016-08-16
Dont requeue stalled jobs

wtfleming:
Too many problems with workers crashing and bee-queue continually re-enqueuing the job, which just makes them crash again.
For now lets keep the crashed jobs in the "stalling" queue instead of automatically moving them back to the waiting queue.


wtfleming:
Sorry, this was meant for a fork, not this repo!

PR #55, opened 2016-12-25
Switch from JSON.parse/stringify to msgpack-lite

v4l3r10:
updated dependencies: switch to last node redis and last dev tool.
switch to msgpack-lite, best performance and memory use (Reduces ~25% the size of job data).
fixed some tests for new redis/msgpack  and eslint

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,4 +1,11 @@
 {
+  "parserOptions": {
+    "ecmaVersion": 6,
+    "sourceType": "module",
+    "ecmaFeatures": {
+        "jsx": true
+    }
+  },
   "rules": {
     "array-bracket-spacing": [2, "never"],
     "brace-style": [2, "1tbs"],
@@ -30,7 +37,7 @@
     "quotes": [2, "single"],
     "semi": [2, "always"],
     "semi-spacing": [2, {"before": false, "after": true}],
-    "space-after-keywords": [2, "always"],
+    "keyword-spacing": [2, {"after": true}],
     "space-before-blocks": 2,
     "space-before-function-paren": [2, {"anonymous": "always", "named": "never"}],
     "space-in-parens": 2,
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ language: node_js
 node_js:
   - "0.12"
   - "0.10"
+  - "4"
 services: redis
 after_script:
   - redis-cli script flush
--- a/lib/job.js
+++ b/lib/job.js
@@ -1,5 +1,6 @@
-var events = require('events');
-var util = require('util');
+const events = require('events');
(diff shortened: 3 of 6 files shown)


thelinuxlich:
Out of curiosity, why the switch to msgpack? Is it better, in this use case, to trade CPU for bandwidth?

matteocontrini:
Also, this would be a breaking change with no compatibility with the data already stored in Redis

v4l3r10:
@thelinuxlich we lose ~10% cpu performance for a reduction of    ~20%/~30% of memory use... and the time for transfer data from redis to node memory... i need to do some benchmark but i think that, overall, Cpu time is the same.

@matteocontrini u never restarted / flushed ur redis database? u leave all data for all time in memory? In my use case i flush redis db every night, so no problem for me.

matteocontrini:
@v4l3r10 what? I have the queue of jobs in the database, why should I even flush it? Data is there to be processed. If this update goes out, jobs that are still to be processed in the database will make the app crash.

v4l3r10:
@matteocontrini REDIS is an in memory database, is not made for  fault-tolerance but only for performance .  I think u need to change Paradigm for your project.

matteocontrini:
Sorry but intentionally flushing the database data is not what I call "fault tolerance" :P

v4l3r10:
No MSGPACK
Ran 10000 jobs through Bee-Queue with concurrency 1 in 4862 ms         3.72MB 
Ran 10000 jobs through Bee-Queue with concurrency 5 in 3496 ms         3.78MB
Ran 10000 jobs through Bee-Queue with concurrency 20 in 3377 ms        3.79MB
Ran 10000 jobs through Bee-Queue with concurrency 50 in 3183 ms        3.79MB
                                                                      
with MSGPACK                                                          
Ran 10000 jobs through Bee-Queue with concurrency 1 in 5502 ms   113%  3.58MB  96%
Ran 10000 jobs through Bee-Queue with concurrency 5 in 4037 ms   115%  3.58MB  94%
Ran 10000 jobs through Bee-Queue with concurrency 20 in 3762 ms  111%  3.64MB  96%
Ran 10000 jobs through Bee-Queue with concurrency 50 in 3627 ms  113%  3.64MB  96%

thelinuxlich:
Sorry, but I fail to see the usefulness of this change in this use case

v4l3r10:
@thelinuxlich  try to work with DateTime Obj with TimeZone.

thelinuxlich:
Care to explain?

v4l3r10:
Try to store a DateTime Obj with timezone in job data. It lose timezone reference during isostring translation. With msgpack i can  preserve all dateTime Obj structure. It's not translated to String, but remain js Obj.

PR #64, opened 2017-07-13
Rewrite bee-queue 1.0.0

skeggse:
~Note that we haven't re-attained 100% code coverage, and that's the biggest blocker to considering this ready to merge.~

- [x] @randallm Implement or remove `start`, `end` parameters for `_getJobs`
- [x] @randallm 100% coverage
- [x] @skeggse Update documentation


HISTORY.md
```diff
@@ -1,14 +1,29 @@
-0.3.0 / 09-01-2015
+1.0.0 / 2017-06-30
```
bradvogel: these new features need to be documented in the readme

package.json
```diff
@@ -28,14 +37,17 @@
...
   ],
+  "engines": {
+    "node": ">= 4"
+  },
   "author": "Lewis J Ellis <me@lewisjellis.com>",
   "license": "MIT",
   "repository": {
     "type": "git",
-    "url": "https://github.com/LewisJEllis/bee-queue.git"
+    "url": "https://github.com/mixmaxhq/bee-queue.git"
```
bradvogel: bee-queue/bee-queue

bradvogel: same with other places in this file

lib/job.js
```diff
@@ -40,78 +46,135 @@ Job.prototype.toData = function () {
...
-      return cb(null, self);
+  const toKey = this.queue.toKey.bind(this.queue);
+
+  let promise;
+  if (this.options.delay) {
+    promise = this.queue._evalScript('addDelayedJob', 4,
+      toKey('id'), toKey('jobs'), toKey('delayed'), toKey('earlierDelayed'),
+      this.id || '', this.toData(), this.options.delay);
+
+    if (this.queue.settings.processDelayed) {
```
bradvogel: setting should be documented in readme

package.json
```diff
@@ -1,22 +1,31 @@
 {
   "name": "bee-queue",
-  "version": "0.3.0",
+  "version": "1.0.0",
   "description": "A simple, fast, robust job/task queue, backed by Redis.",
   "main": "index.js",
   "dependencies": {
-    "redis": "1.0.0"
+    "promise-callbacks": "^3.0.0",
```
LewisJEllis: I'd much prefer if we pin these. I'm guessing it's just oversight, but I'm happy to discuss if you'd explicitly rather not.

skeggse: We prefer to not pin dependencies, because it seems more the job of the application to ensure that its dependencies are locked. In our case, we use `yarn`, but other folks may use npm shrinkwrap or lockfiles. For us, not pinning dependencies is beneficial because [yarn deduplicates common dependencies](https://gist.github.com/vvo/84a94cfc0f94c91ea6b6#pinning-dependencies-will-result-in-duplicated-code-and-modules-in-builds).

package.json
skeggse: Oops, thought I got these all. Done.

LewisJEllis:
My 10-minute skim impression/thoughts:
- Looks great overall, you've kept along a lot of the original ideas/tenets when starting this and polished it to be 2017 instead of 2015. Most of the spots where the 2-years-later version of me sees room for stuff to be better, it is, especially around tooling/CI/etc
- Good move on the promise transition. I skipped them previously to dodge bluebird/a compilation step/node 0.12's sketchy promise situation, but now that they're a standard language feature (along with async/await) I'm happy to see/use them. Curious about your thoughts on the option of moving to an entirely promise-first API and maybe obviating need for `promise-callback`?
- I'm gonna have to go through in more detail to understand the new timer stuff and the new behavior for stalling/delayed jobs; definitely adds complexity, but it looks well-thought-out.

It sounds like you're still working on tests, which do make review easier, so I'm inclined to wait on those, but just let me know whenever you think they're are in good shape for review. Maybe that's even now, I'm not sure - happy to review now during WIP or wait, just let me know either way. Just hoping to avoid leaving comments of "there should be a test for this" when you already know :)

LewisJEllis:
Oh, also definitely looking for more readme updating, especially if you want to go waving a flag of "hey we're updated and back in action". I'd guess my initial release time breakdown was 1:1:1 impl/tests/readme, and I think that was a significant factor to why this initially caught some adoption (well, that and the whim of the editor of the Node Weekly newsletter...).

bradvogel:
@LewisJEllis fyi: we're still debugging one last issue with this PR (running this in production!). So wait on reviewing.

skeggse:
@LewisJEllis it looks like maybe Coveralls isn't functioning now that bee-queue moved to its own organization. I'm not able to get Coveralls to discover bee-queue via my account, can you take a look?

LewisJEllis:
Sure thing - I went into my coveralls account and hit some sort of button that turned https://coveralls.io/github/lewisjellis/bee-queue into https://coveralls.io/github/bee-queue/bee-queue and synced/updated everything; I think it should try to do a report on the next push, so give it a try. If that doesn't work, you also might try moving to a new PR that's on a branch in this repo rather than on the mixmaxhq fork.

skeggse:
We're running into what may be a performance regression Node-side (Redis is still performing quite nicely) - I suspect that the switch to promises, especially ES6 promises, may be creating substantially more objects than we expected. I'm seeing GC overhead of up to 15% in production. Thoughts about performance and promises, @LewisJEllis? It looks like maybe bluebird is more optimized relative to es6 promises, so maybe we could support overriding the `Promise` implementation at runtime.

LewisJEllis:
That's actually entirely possible. I recall playing with Bluebird's benchmarks during initial dev here and deciding the the speed advantage of raw callbacks *could* be significant in extremely high-throughput scenarios, but I kind of assumed serialization and IO would still dominate and so I focused most of my efforts on minimizing roundtrips and maximizing pipelining & effective use of Lua scripting.

It sounds like that might not be the case, though; I just ran some of bluebird's benchmarks:
```
results for 50000 parallel executions, 1 ms per I/O op

file                                       time(ms)  memory(MB)
callbacks-baseline.js                          1110      115.84
promises-bluebird-generator.js                 1940      209.61
promises-bluebird.js                           2332      220.53
promises-ecmascript6-native.js                 6899      430.85

Platform info:
Darwin 16.6.0 x64
Node.JS 8.1.2
V8 5.8.283.41
Intel(R) Core(TM) i5-6267U CPU @ 2.90GHz × 4
```
and at 50 parallel:
```
results for 10000 parallel executions, 1 ms per I/O op

file                                      time(ms)  memory(MB)
callbacks-baseline.js                          776      136.92
promises-bluebird.js                           999      181.27
promises-bluebird-generator.js                1012      187.42
promises-ecmascript6-native.js                5035      574.09

Platform info:
Darwin 16.6.0 x64
Node.JS 8.1.2
V8 5.8.283.41
Intel(R) Core(TM) i5-6267U CPU @ 2.90GHz × 4
```
It looks like native promises are 5-6x slower than raw callbacks and 3-5x slower than bluebird here. I don't think the delta between callbacks and bluebird promises would be significant, but maybe with native promises it would be. Definitely worth benchmarking, and ultimately I think using bluebird could be reasonable for performance reasons.

skeggse:
I ran a battery of benchmarks against different node versions, and three variations of bee queue (alongside bull and kue) - I'll report back with my findings once I finish digging through them.

skeggse:
> Curious about your thoughts on the option of moving to an entirely promise-first API and maybe obviating the need for `promise-callbacks`?

We use other bits from `promise-callbacks`. At the very least, I think it makes sense to `wrapAsync` around the job handler to allow users to either `return` a `Promise` or call a callback. It also provides `waitOn` for simple `Promise`-oriented `EventEmitter#once` functionality, `withTimeout` for `Job` timeouts, and `deferred` for interop with the callback-oriented redis commands.

It seems (relatively) painless to include `if (cb) helpers.asCallback(promise, cb);` in all the async methods, and enables better backwards compatibility to reduce the migration folks may need to do between versions. I'm open to being swayed here, yet it seems more difficult to me to retrofit a promise-returning API with callbacks externally than it does to use `promisify` going the other direction.

skeggse:
Getting close to being ready - just waiting on a couple tasks from @randallm.

@LewisJEllis, just to follow up on some prior convo, and start some new:

- Happy to pin `promise-callbacks` if you'd like, but I'd prefer leaving it unpinned.
- Do you have any preference on leaving in the `verbose` event for external logging? We found it necessary to debug a couple of issues, but it does add a bit of extra code, and it can be somewhat brittle in rare error handlers.
- Any feedback on how the tests read? I think they're more readable than with the callback pasta (it wasn't quite spaghetti :D), but they still have enough interdependencies and strange synchronization mechanisms that I could see them being pretty hard to get used to.

Thanks for your support in this endeavor!

lib/lua/checkStalledJobs.lua
```diff
@@ -13,30 +12,38 @@ if a jobId is not removed from the stalling set within a stallInterval window,
...
-  redis.call("rpush", KEYS[3], unpack(stalling))
-  for i = 1, #stalling do
-    redis.call("lrem", KEYS[4], 0, stalling[i])
+local stalling, stalled = redis.call("smembers", KEYS[2]), {}
+if next(stalling) ~= nil then
+  -- not worth optimizing - this should be a rare occurrence, better to keep it straightforward
+  local nextIndex = 1
+  for i, jobId in ipairs(stalling) do
+    local removed = redis.call("lrem", KEYS[4], 0, jobId)
+    -- we only restart stalled jobs if we can find them in the active list - otherwise, the stalled
```
bradvogel: this theoretically can't happen, right?

skeggse: best I figure, yeah. I updated the comment

lib/defaults.js
```diff
@@ -1,9 +1,31 @@
...
+  ensureScripts: true,
+  processDelayed: false,
   sendEvents: true,
+  storeJobs: true,
   removeOnSuccess: false,
-  catchExceptions: false
+  removeOnFailure: false,
+  catchExceptions: false,
+  redisScanCount: 100,
+  verbose: false,
```
bradvogel: is this used?

lib/queue.js
```diff
@@ -1,256 +1,527 @@
...
+
+  if (cb) helpers.asCallback(promise, cb);
+  return promise;
+};
+
+Queue.prototype._waitForJob = function () {
+  const idPromise = helpers.deferred(), start = Date.now();
+  let getStart;
+  /* istanbul ignore next */
+  this._verbose('blockingPop');
```
bradvogel: do we still need all these? Seems excessive and was only used once for debugging redis

LewisJEllis:
@skeggse I took more of a read over this yesterday and am reading more/writing up thoughts this morning; here's some initial bits before I get into the code:

- re: pin vs no-pin: yea, the "ensure versions are locked for reproducible builds" concern has been well-addressed by yarn and npm5, but I personally still pin primarily from a security perspective; wouldn't want a malicious patch release of a dependency to affect bee-queue users. Of course, if someone starts going around publishing malicious patch releases, it's likely that many bee-queue users would still fall victim via some other dependency and bee-queue being paranoid wouldn't have helped. Plus, node-redis is the super-widely-used node-redis and callback-promise is a mixmax package, so it's also theoretically more trusted - but I do think it's kind of strange/borderline-concerning to have 18 accounts with publishing rights on it (that's probably a separate discussion though). As long as that malicious-patch-versions concern is well-understood, I'm fine with not pinning.

- re: how tests read: I like this beforeEach context setup pattern and I really like using async/await in the tests. Seems good from what I've read so far, will keep a mental note if any of the nontrivial synchronization situations you mention start to feel weird as I get into the tests more.

- re: verbose events: if you think they're likely to be useful again in the future, it seems reasonable to keep them, but if they were only ever used to debug redis once (that pre-redis-3.2 problem I'm guessing?) then I'd rather get rid of them. It'd be nifty if it could be a build-step thing where they get pulled out by a dead-code-elimination type thing, but we don't even *have* a build step so that's probably more effort than it's worth.

- what was your final take on including bluebird to replace native promises? seems worth doing afaict, and I don't mind the additional dependency if it makes an appreciable performance improvement

- Brad mentioned a possible Redis 2.8 bug that came up, curious to hear more details on that

- general disclaimer: I'm likely to leave a couple performance nitpick comments on differences that would probably only be measurable with node isolated and not measurable in an actual benchmark with actual redis I/O going on, so feel free to take them with a grain of salt

lib/queue.js
skeggse: Gonna remove them.

skeggse:
* module pinning: I'm filing an issue internally to set up additional controls for that kind of thing - require signoff, and publish via travis
* verbose events: I'm in favor of removing them.
* promises: I'd like to have a pattern where the `Promise` implementation is specified by the user; users who would rather not have the additional dependency can just use the module in its default configuration, and users who need the extra performance can configure that per-queue (thoughts? either in the `options` or via `queue.setPromiseImplementation(Promise)` or something). I think I'll submit a separate pull-request for that, because the changes are mostly independent to this PR.
* the Redis 2.8 bug appeared to involve Redis not properly handling the `BRPOPLPUSH` command, where long-running invocations would somehow stall, and would execute the RPOPLPUSH but not pass the data to the client. This resulted in additional stalled jobs where the consumer was perfectly healthy.

LewisJEllis:
Sounds good, sounds good, re: promises, sounds good to have an `options` field to let a user pass in their bluebird etc. I'd be curious to see if/how that affects benchmarks.

Interesting on the Redis 2.8 bit; some of my usage of this probably had brpoplpush running for weeks at a time between receiving jobs and I never noticed anything like that. You guys probably run more jobs through this in a week than I ever have, though.

Still working on review...

skeggse:
It'd be super awesome to have automated benchmarks set up - it's currently really time-consuming to run the benchmarks :(

LewisJEllis:
Would definitely be nice to have it automatically tied into CI, but I'm not sure if benchmarking on travis is the best idea. Maybe make a fire-and-forget `run-benchmarks.sh` that goes through all cases, to ease the manual-benchmark-run process? I can do that as a follow-on and we can update the charts later if it turns out bluebird has a significant effect.

lib/queue.js
bradvogel: thanks - yeah, prefer to keep it clean

lib/queue.js
```diff
@@ -1,256 +1,497 @@
...
+    });
   }
-  this.settings.redis.db = this.settings.redis.db || 0;
 
-  // Wait for Lua loading and client connection; bclient and eclient/subscribe if needed
-  var reportReady = barrier(
-    2 + this.settings.isWorker + this.settings.getEvents * 2,
-    this.emit.bind(this, 'ready')
-  );
+  this._delayedTimer = new EagerTimer(this.settings.nearTermWindow);
```
LewisJEllis: I guess this is fine behavior-wise since `_processDelayed` early-exits if `options.processDelayed` is false, but it seems weird to create this timer if we know it'll never do anything. The only place that touches the timer when `options.processDelayed` is false is the `close()` cleanup stopping it, so I think it would be reasonable to only create the timer if `options.processDelayed` is true.

README.md
```diff
@@ -193,325 +243,482 @@ To make this happen, workers periodically phone home to Redis about each job the
...
-- `sendEvents`: boolean, default true. Disable if this worker does not need to send job events back to other queues.
-- `removeOnSuccess`: boolean, default false. Enable to have this worker automatically remove its successfully completed jobs from Redis, so as to keep memory usage down.
-- `catchExceptions`: boolean, default false. Only enable if you want exceptions thrown by the [handler](#queueprototypeprocessconcurrency-handlerjob-done) to be caught by Bee-Queue and interpreted as job failures. Communicating failures via `done(err)` is preferred.
+- `isWorker`: boolean. Disable if this queue will not process jobs.
+- `getEvents`: boolean. Disable if this queue does not need to receive job events.
+- `sendEvents`: boolean. Disable if this worker does not need to send job events back to other queues.
+- `storeJobs`: boolean. Disable if this worker does not need to associate events with specific `Job` instances.
+- `ensureScripts`: boolean. Ensure that the LUA scripts exist in redis before running any commands against redis.
+- `processDelayed`: boolean. Activate delayed jobs once they've passed their `delayUntil` timestamp.
```
LewisJEllis: Thoughts on calling this something like `raiseDelayedJobs` or `promoteDelayedJobs` instead? A worker queue having `processDelayed` false might erroneously suggest to an uninformed reader that that worker will not process jobs that were delayed.

lib/queue.js
```diff
@@ -1,256 +1,497 @@
...
 
 Queue.prototype.destroy = function (cb) {
-  cb = cb || helpers.defaultCb;
-  var keys = ['id', 'jobs', 'stallTime', 'stalling', 'waiting', 'active', 'succeeded', 'failed']
-    .map(this.toKey.bind(this));
-  this.client.del.apply(this.client, keys.concat(cb));
+  const promise = this._commandable().then((client) => {
+    const deleted = helpers.deferred();
+    const args = ['id', 'jobs', 'stallBlock', 'stalling', 'waiting', 'active',
+      'succeeded', 'failed']
```
LewisJEllis: add `delayed` here?

lib/queue.js
```diff
@@ -1,256 +1,497 @@
...
+    // jobs while waiting for the ongoing jobs to terminate.
+    if (this.settings.isWorker) {
+      this.bclient.end(true);
+    }
+
+    // Wait for all the jobs to complete. Ignore job errors during shutdown.
+    const waitJobs = Array.from(this.activeJobs);
+    return Promise.all(waitJobs.map((promise) => promise.catch(() => {})));
+  }), timeout).then(() => {
+    return cleanup().then(() => undefined);
```
LewisJEllis: I almost want to say `=> void 0` or `=> {}` here, but that's paranoid-level defensive programming. Reads better like this.

lib/queue.js
```diff
@@ -1,256 +1,497 @@
...
+      default:
+        throw new Error('Improper queue type');
+    }
+
+    return idsPromise;
+  }).then((ids) => {
+    // If ids is a Set (i.e. returned from Queue._scanForJobs), convert it to an
+    // Array.
+    const idArray = Array.from(ids);
+    return Promise.all(idArray.map((id) => this.getJob(id)));
```
LewisJEllis: This will end up doing a bunch of `hget`s in `fromId` for jobs that aren't stored in the local jobs map; we *could* optimize by doing an `hmget` if we were concerned about getJobs performance. Since we don't use it anywhere else inside bee-queue and it's mainly provided for inspection by e.g. arena, I'm not too worried about this, but with large pages of delayed jobs, it might be an appreciable performance difference.

lib/eager-timer.js
```diff
@@ -0,0 +1,85 @@
...
+    this._stop();
+    this._nextTime = time;
+    this._timer = setTimeout(this._boundTrigger, duration);
+  }
+
+  _trigger() {
+    const now = Date.now(), remaining = this._nextTime - now;
+    /* istanbul ignore if */
+    if (remaining > 0) {
+      // It's possible (it caused tests to fail) to have the timeout trigger
```
LewisJEllis: curious if the specific cases where that happens are known? or is it more of a "well it was messing up somehow and this fixes it"?

lib/queue.js
```diff
@@ -259,104 +500,199 @@ Queue.prototype.finishJob = function (err, data, job, cb) {
...
-        return setImmediate(jobTick);
+    // invariant: in this code path, this.running < this.concurrency, always
+    // after spoolup, this.running + this.queued === this.concurrency
+    this._getNextJob().then((job) => {
+      // We're shutting down.
+      if (this.paused) {
+        // This job will get picked up later as a stalled job if we happen to
+        // get here. We can't easily process this job because at this point
+        // Queue#close has already captured the activeJobs set in a Promise.all
+        // call, and we'd prefer to delay a job than half-process it.
```
LewisJEllis: nice call on this edge case

lib/queue.js
```diff
@@ -1,256 +1,527 @@
...
+const helpers = require('./helpers');
+const backoff = require('./backoff');
+const EagerTimer = require('./eager-timer');
 
 function Queue(name, settings) {
   if (!(this instanceof Queue)) {
     return new Queue(name, settings);
   }
 
+  Emitter.call(this);
```
skeggse: When I still naïvely expected this would be a small PR, I decided against switching to `class`es because it would increase the diff size. I like `class`es. I will do this. :)

lib/queue.js
skeggse: Clearly _that_ test is incomplete >.<

lib/queue.js
```diff
@@ -1,256 +1,497 @@
...
     });
   };
   preventStalling();
 
-  var handleOutcome = function (err, data) {
-    // silently ignore any multiple calls
-    if (handled) {
-      return;
+  const handleOutcome = (err, data) => {
+    completed = true;
```
skeggse: In this case, it isn't necessary because we moved to promises. The promise cannot `resolve` and/or `reject` twice, so this function will be called at most once.

lib/queue.js
skeggse: Nothing like production to help you find edge cases 😝

lib/queue.js
```diff
@@ -259,104 +500,199 @@ Queue.prototype.finishJob = function (err, data, job, cb) {
...
+  this._processDelayed();
 
-  this.checkStalledJobs(setImmediate.bind(null, jobTick));
+  return this;
 };
 
-Queue.prototype.checkStalledJobs = function (interval, cb) {
-  var self = this;
-  cb = typeof interval === 'function' ? interval : cb || helpers.defaultCb;
+Queue.prototype._doStalledJobCheck = function () {
```
skeggse: This separation actually reduced complexity of `_checkStalledJobs` for me - I'd prefer to keep it as-is. They represent distinct functions, so I think it makes sense to leave them as distinct _functions_ :p

lib/queue.js
```diff
@@ -259,104 +530,217 @@ Queue.prototype.finishJob = function (err, data, job, cb) {
...
   return this.settings.keyPrefix + str;
 };
 
+/**
+ * Evaluate the named script, return a promise with its results.
+ *
+ * Same parameter list/syntax as evalsha, except for the name.
+ */
+Queue.prototype._evalScript = function (name) {
+  const args = helpers.toArray(arguments);
```
skeggse: TIL this is a thing, but it makes a lot of sense. Thanks!

Gruntfile.js
```diff
@@ -1,26 +1,17 @@
 module.exports = function(grunt) {
```
skeggse: Awesome, not a huge fan of grunt anyway 👍

lib/lua/addDelayedJob.lua
```diff
@@ -0,0 +1,30 @@
...
+key 2 -> bq:name:jobs
+key 3 -> bq:name:delayed
+key 4 -> bq:name:earlierDelayed
+arg 1 -> job id
+arg 2 -> job data
+arg 3 -> job delay timestamp
+]]
+
+local jobId = ARGV[1]
+if jobId == "" then
```
skeggse: I thought I had a reason for doing this, but it must have been refactored >.<

lib/lua/addDelayedJob.lua
```diff
@@ -0,0 +1,30 @@
...
+end
+redis.call("hset", KEYS[2], jobId, ARGV[2])
+redis.call("zadd", KEYS[3], tonumber(ARGV[3]), jobId)
+
+-- if this job is the new head, alert the workers that they need to update their timers
+-- if we try to do something tricky like checking the delta between this job and the next job, we
+-- can enter a pathological case where jobs incrementally creep sooner, and each one never updates
+-- the timers
+local head = redis.call("zrange", KEYS[3], 0, 0)
+if #head > 0 and head[1] == jobId then
```
skeggse: ah true

lib/lua/checkStalledJobs.lua
```diff
@@ -13,30 +12,38 @@ if a jobId is not removed from the stalling set within a stallInterval window,
...
+if next(stalling) ~= nil then
+  -- not worth optimizing - this should be a rare occurrence, better to keep it straightforward
+  local nextIndex = 1
+  for i, jobId in ipairs(stalling) do
+    local removed = redis.call("lrem", KEYS[4], 0, jobId)
+    -- safety belts: we only restart stalled jobs if we can find them in the active list
+    -- the only place we add jobs to the stalling set is in this script, and the two places we
+    -- remove jobs from the active list are in this script, and in the MULTI after the job finishes
+    if removed > 0 then
+      stalled[nextIndex] = jobId
```
skeggse: true, but the `#` operator is [more expensive](https://stackoverflow.com/a/18208085) - I'd rather this be O(n) than O(nlogn).

skeggse: That said, I did literally leave a comment that says `-- not worth optimizing - this should be a rare occurrence, better to keep it straightforward`

what do you think?

lib/lua/checkStalledJobs.lua
```diff
@@ -13,30 +12,38 @@ if a jobId is not removed from the stalling set within a stallInterval window,
...
+      nextIndex = nextIndex + 1
+    end
+  end
+  -- don't lpush zero jobs (the redis command will fail)
+  if nextIndex > 1 then
+    redis.call("lpush", KEYS[3], unpack(stalled))
```
skeggse: This is just to avoid stalled jobs hogging the consumers if for some reason a specific job causes an uncaught exception. I'll add an explanatory comment.

package.json
```diff
@@ -1,22 +1,32 @@
...
-    "coveralls": "^2.11.2"
+    "lolex": "^2.0.0",
+    "nyc": "^11.0.3",
+    "sandboxed-module": "^2.0.3",
+    "sinon": "^2.3.8"
   },
   "scripts": {
-    "test": "grunt test"
+    "test": "npm run eslint && ava",
+    "eslint": "grunt eslint",
```
skeggse: I'm gonna do `eslint lib/ test/` unless you have a specific reason not to

README.md
skeggse: Good point 👍

README.md
```diff
@@ -193,325 +243,482 @@ To make this happen, workers periodically phone home to Redis about each job the
...
+- `removeOnSuccess`: boolean. Enable to have this worker automatically remove its successfully completed jobs from Redis, so as to keep memory usage down.
+- `removeOnFailure`: boolean. Enable to have this worker automatically remove its failed jobs from Redis, so as to keep memory usage down. This will not remove jobs that are set to retry unless they fail all their retries.
+- `catchExceptions`: boolean. Only enable if you want exceptions thrown by the [handler](#queueprototypeprocessconcurrency-handlerjob-done) to be caught by Bee-Queue and interpreted as job failures. Communicating failures via `done(err)` is preferred.
```
skeggse: On the one hand, I feel like this has value for when you're misusing another library, and it throws a synchronous exception - then you get an `error` event instead of potentially retrying the job. On the other hand, `async` functions will reject the promise instead of throwing a synchronous exception. I think this makes sense, and folks will just need to deal with misconfigurations in some better manner.

lib/eager-timer.js
skeggse: It was more of a "this causes a race condition, so it's hard to write a test that reproduces it" situation

lib/queue.js
skeggse: might as well

lib/queue.js
```diff
@@ -45,8 +45,13 @@ class Queue extends Emitter {
       });
     }
 
-    this._delayedTimer = new EagerTimer(this.settings.nearTermWindow);
-    this._delayedTimer.on('trigger', this._activateDelayed.bind(this));
+    // To avoid changing the hidden class of the Queue.
+    this._delayedTimer = this.settings.activateDelayedJobs
```
LewisJEllis: I think this could just be
```js
if (this.settings.activateDelayedJobs) {
  this._delayedTimer = new EagerTimer(this.settings.nearTermWindow);
  this._delayedTimer.on('trigger', this._activateDelayed.bind(this));
}
```
The `if (this._delayedTimer) {` check will still work the same way with _delayedTimer just being undefined instead of null.

lib/lua/checkStalledJobs.lua
LewisJEllis: I think it's not worth optimizing; `n` is in almost all cases bounded by the maximum processing concurrency of a consumer, which is unlikely to be terribly large.

lib/queue.js
LewisJEllis: Yea, I wouldn't mind at all if it was just a *function*, but since it's a *method* it feels sort of extraneous. Not a big deal, it's fine as is.

lib/queue.js
```diff
@@ -276,6 +276,37 @@ class Queue extends Emitter {
...
+    const idsArray = Array.from(ids);
+    if (!idsArray.length) {
+      return Promise.resolve([]);
+    }
+
+    // We need to re-ensure the queue is commandable, as we might be shutting
+    // down during this operation.
+    return this._commandable().then((client) => {
+      const got = helpers.deferred();
+      const commandArgs = [this.toKey('jobs')].concat(idsArray, got.defer());
```
LewisJEllis: In the previous getJobs we would just use the job instance from the local jobs map if available; here it seems we fetch all jobs from redis and make new instances. Could we continue to use the local instances that are available and only `hmget` the missing ones?

lib/queue.js
skeggse: Yeah, it'd work. I prefer to keep properties on objects in all cases, because doing otherwise introduces a different hidden class. I don't expect that to really have any performance benefits. I also think it's nice to always have the full set of properties always present on an object, so you can see them easily in the REPL.

lib/lua/checkStalledJobs.lua
skeggse: 👍

lib/queue.js
LewisJEllis: Yea, I'm also doubtful it would have any performance benefit since it's just 1 vs 2, not 1 vs 20 or log(n) or n - but the "always see all properties in REPL" bit is a fair point that I hadn't considered. Sounds good as is.

lib/queue.js
LewisJEllis: My specific thinking here is that this could have some benefit in the use case of "arena is sitting here calling getJobs() every 10 seconds to keep an up-to-date dashboard display of jobs that have failed". I have no idea if arena in particular actually does something like that or if the intent is to use it like that, but it seems like a reasonable monitoring possibility.

LewisJEllis: Although on second thought, maybe in that case we would want to have disabled `storeJobs` on that queue instance anyway to avoid memory issues, and so this wouldn't make any difference. I guess the "watch jobs that have failed" use case wouldn't be so sensitive to it assuming we don't end up with many failed jobs, but for the "watch jobs that are waiting" use case we would almost certainly want to have `storeJobs` disabled.

skeggse: [Related](https://github.com/bee-queue/arena/pull/38). I'm adding this functionality, but it won't be used in Arena for them time being.

skeggse:
🎉

LewisJEllis:
Added a .npmignore in a8e7183 and published!

skeggse:
The other option there, @LewisJEllis, is to specify a whitelist via the `files` array in the `package.json`.

LewisJEllis:
Oh, nice, TIL. That seems way easier given that I started with a whitelist in the comments of the .npmignore.

Issue #16, opened 2015-09-24
Replace node_redis with ioredis


thelinuxlich:
For redis 3 cluster support

LewisJEllis:
I've had my eye on ioredis and will give this some consideration. When I started on this library it hadn't been released yet, but now it's looking quite nice.

thelinuxlich:
I forked bee-queue to add ioredis and only two tests are failing(reconnection)

LewisJEllis:
Cool! Off the top of my head, I would expect those two tests to fail among all others; they're pretty specific to how node_redis deals with connections. I made an ioredis branch; feel free to open a PR to it and we can work from there.

thelinuxlich:
Actually there were a lot of tests failing and now there are only two: Queue "after each" hook and Queue Health Check reports a failed job

My fork https://github.com/thelinuxlich/bee-queue/commit/795a465ac4fa047c3ed0b8ba831129c45bb84288

LewisJEllis:
Moving forward with this in #21. I just want to leave a note here on cluster support in general; 

Because of the nature of Redis and its lists and (mostly) its Lua scripting, bee-queue won't exactly be able to offer true cluster support. It will work on Redis cluster deployments, but it will really just run everything on one Redis instance. Otherwise, the Lua scripts won't be able to access all the keys they need to in one place.

Fortunately [Redis cluster]() has a very nice way to do this:

> Redis Cluster supports multiple key operations as long as all the keys involved into a single command execution (or whole transaction, or Lua script execution) all belong to the same hash slot. The user can force multiple keys to be part of the same hash slot by using a concept called hash tags.
>
> Hash tags are documented in the Redis Cluster specification, but the gist is that if there is a substring between {} brackets in a key, only what is inside the string is hashed, so for example this{foo}key and another{foo}key are guaranteed to be in the same hash slot, and can be used together in a command with multiple keys as arguments.

but it should be clear that any one instance of bee-queue will still be running entirely on one instance of Redis. Of course, it will be possible to run multiple instances of bee-queue on multiple instances of Redis.

hustcer:
+1
need a update on this, asap. Thanks.

skeggse:
From #21 
[...]

PR #68, opened 2017-08-11
Support passing redis client in settings object

skeggse:
This allows client sharing between multiple Queue instances, or other uses.

It also begins to add support for ioredis (cc #16 and #21). We might consider supporting both, though there are some major inconsistencies between the two libraries, such as the multi-exec callback form.

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -30,6 +30,7 @@ class Queue extends Emitter {
     settings = settings || {};
     this.settings = {
       redis: settings.redis || {},
+      quitCommandClient: settings.quitCommandClient,
       keyPrefix: (settings.prefix || defaults.prefix) + ':' + this.name + ':'
     };
 
@@ -49,6 +50,12 @@ class Queue extends Emitter {
       });
     }
 
+    // By default, if we're given a redis client and no additional instructions,
+    // don't quit the connection on Queue#close.
+    if (typeof this.settings.quitCommandClient !== 'boolean') {
+      this.settings.quitCommandClient = !redis.isClient(this.settings.redis);
+    }
+
     // To avoid changing the hidden class of the Queue.
     this._delayedTimer = this.settings.activateDelayedJobs
       ? new EagerTimer(this.settings.nearTermWindow)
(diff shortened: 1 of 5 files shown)


lib/redis.js
```diff
@@ -3,17 +3,70 @@
...
+function isAbortError(err) {
+  // node_redis has a designated class for abort errors, but ioredis just has
+  // a constant message defined in a utils file.
+  return err.name === 'AbortError' ||
+    /* istanbul ignore next: this is only for ioredis */
+    err.message === 'Connection is closed.';
+}
+
+function isClient(object) {
+  if (!object || typeof object !== 'object') return false;
```
LewisJEllis: no gotcha here on the fact that `typeof null === 'object'`, checks out

lib/queue.js
```diff
@@ -184,7 +191,10 @@ class Queue extends Emitter {
     const cleanup = () => {
       this._isClosed = true;
 
-      const clients = [this.client];
+      const clients = [];
+      if (this.settings.quitCommandClient) {
```
LewisJEllis: semi-uninformed question, I think it's fine as is and I probably just need reminding: on line 204 we do `client.quit()` with each of these; should we keep doing that or would we want to use new `disconnect()` helper? I don't recall exact differences of `end()` vs `quit()` and why we use which one where.

test/queue-test.js
```diff
@@ -370,6 +378,31 @@ describe('Queue', (it) => {
...
+
+        t.true(client.ready);
+        t.false(client.quit.called);
+
+        queue = t.context.makeQueue({
+          redis: client,
+          quitCommandClient: true
+        });
+
+        await queue.close();
```
LewisJEllis: do we test (or care about) the case where we pass `quitCommandClient: false` but don't also provide an existing redis client?

lib/queue.js
skeggse: We still want to use `quit`, as it tells Redis to disconnect, and ensures we flush our command queue in the process. We can't use `quit` with `bclient` because chances are it's waiting for a job, with no timeout.

test/queue-test.js
skeggse: Yep! 70dcb515c6a9e9c3fd2c712b53ed94ae98ecd0a9

Issue #67, opened 2017-08-09
Add Redis resource utilization benchmarks


skeggse:
We updated the benchmarks to show throughput for job processing across different job queues, but didn't instrument or compare their Redis utilization. We should do that.

manast:
I recommend you do that after this PR is merged: https://github.com/OptimalBits/bull/pull/641

skeggse:
Some related results: #69. Unfortunately not a complete Redis util benchmark.

bradvogel:
Thanks for adding. So it looks like based on the raw results, Bee Queue is still roughly 2x faster than Bull 3.0rc5. Cc @manast

manast:
@bradvogel hmm, more like 1.5x I would say. I am quite happy with the results so far. The concurrency issue is gone which was the most important issue in 3.0.

manast:
Btw. something to consider. Bull has many more .lua files than bq to read and compile when it starts. This time is also computed in the test, could add 0.5 to 1 second to the total benchmark time.

bradvogel:
Youre right - about 1.5x. Not sure about the startup time, but that could have also influenced the tests. Looks like Bull 3rc5 is getting close, and has just fine performance for 99% of queue use cases.

PR #69, opened 2017-08-12
Update the benchmark results and summary

skeggse:
* Added a benchmark script (be careful with it).
* Added a silly analysis script (improvements welcome).
* Consolidated the actual tests.
* Published the raw results from July.

Doesn't quite get us #67, because it looks like my `aws-benchmark.sh` script didn't actually capture all the Redis stats I wanted.


bradvogel:
@LewisJEllis updated the readme

README.md
```diff
@@ -96,21 +96,9 @@ Bee-Queue is like a bee because it:
 
 # Benchmarks
 
-![benchmark chart](https://raw.githubusercontent.com/bee-queue/bee-queue/master/benchmark/benchmark-chart.png)
+![benchmark chart](./benchmark/results-2017-08-12/resultschart.png)
```
skeggse: This will make the image not load on npmjs.com - what's the benefit to doing this?

benchmark/results-2017-08-12/resultschart.html
```diff
@@ -0,0 +1,83 @@
+<!DOCTYPE HTML>
```
skeggse: Thanks!

README.md
```diff
@@ -96,21 +96,9 @@ Bee-Queue is like a bee because it:
...
-
-| Library   | Memory After  | Memory Peak | Memory After Δ | Memory Peak Δ |
-| --------- | ------------- | ----------- | -------------- | ------------- |
-| Bee-Queue |        2.65MB |      2.73MB |         1.67MB |        1.75MB |
-| Bull      |        3.93MB |      5.09MB |         2.95MB |        4.11MB |
-| Kue       |        7.53MB |      7.86MB |         6.55MB |        6.88MB |
-
-The Δ columns factor out the ~986KB of memory usage reported by Redis on a fresh startup. Note that this data was collected in 2015, but should still accurately reflect the memory usage of Bee-Queue.
+These basic benchmarks ran 10,000 jobs through each library, at varying levels of concurrency, with Node.js (v6.9.1, v6.11.2, v7.6.0, v7.10.1, v8.2.1, v8.3.0) and Redis (v3.2.10, v4.0.1) running directly on an AWS m4.xlarge. The numbers shown are averages of 3 runs; the raw data collected and code used are available in the [benchmark folder](./benchmark).
```
skeggse: Ah I forgot I changed the instance size. I'll update.

README.md
bradvogel: Oh, I'll fix. That's lame of npm :-P

README.md
skeggse: The `./benchmark` link won't work on npm

benchmark/bull/harness.js
```diff
@@ -1,26 +1,50 @@
...
+      return true;
     }
   };
+}
 
-  queue.process(options.concurrency, function (job, done) {
-    reportResult();
-    return done();
-  });
+module.exports = (options) => {
```
LewisJEllis: Can we wrap the body of this in a `queue.isReady().then(...)`? I believe that would address the concern here about startup costs/Lua script caching time: https://github.com/bee-queue/bee-queue/issues/67#issuecomment-322008724

benchmark/bq/harness.js
```diff
@@ -1,26 +1,57 @@
...
+      return true;
     }
   };
+}
 
-  queue.process(options.concurrency, function (job, done) {
-    reportResult();
-    return done();
-  });
+module.exports = (options) => {
```
LewisJEllis: Let's similarly wrap this in a `queue.ready().then(...)` to factor out startup time.

benchmark/results-2017-08-12/summary-2017-08-12.txt
```diff
@@ -0,0 +1,125 @@
...
+    bq-min => [2690 :: 4553 {median: 3016, mean: 3286.4}]
+      bull => [4676 :: 7408 {median: 5584, mean: 5673.7}]
+       kue => [8673 :: 13511 {median: 9907, mean: 10375.7}]
+ 7.6.0 =>
+        bq => [2712 :: 4593 {median: 3210, mean: 3345.0}]
+      bq-0 => [2714 :: 4603 {median: 3234, mean: 3354.4}]
+    bq-min => [2550 :: 4273 {median: 2894, mean: 3096.5}]
+      bull => [4954 :: 7446 {median: 5827, mean: 5856.7}]
+       kue => [8338 :: 12680 {median: 9324, mean: 9736.9}]
+7.10.1 =>
```
LewisJEllis: If I'm interpreting these numbers correctly, it looks like bee-queue is significantly faster on Node 7 and even 6 than on 8, while others are more consistent across Node versions. Any idea why that might be? Is it worth pointing out in the readme?

bradvogel: (filling in for @skeggse, as he's out of the country for several months) I'm not sure why this is.  Though I'm curious to rerun it on Node 8.4 with the new, faster v8.

benchmark/bq/harness.js
bradvogel: Fixed

bradvogel: Fixed

benchmark/bull/harness.js
bradvogel: Fixed

bradvogel:
@LewisJEllis made changes - this is ready for re-review

bradvogel:
@LewisJEllis can we merge?

manast:
Bull 3.0 final has been released. This release has squeezed a bit more of performance, so you may be interested in update the benchmarks with it.

bradvogel:
@manast Re-ran with Bull 3.0: https://github.com/bee-queue/bee-queue/pull/69/commits/5de983b5cf1dafe3aec088f9e269dfdf9e95fd4e. The machine I ran it on must have been slightly slower (it was also an m4.large instance, so not sure why), because all the numbers were slightly slower.

manast:
@bradvogel did you also run bq benchmark in the same machine?

manast:
@bradvogel I wonder, have you made any benchmarks with several processes and/or machines running in parallel? I assume in current benchmark its the machines CPU (at least one core) and the IO which are the bottlenecks. It would be quite interesting to know what is the limit per redis instance as well.

manast:
I hacked cluster support in bull harness, I am seeing a nice performance improvement however less than logarithmic, just with 2 workers (one process for adding jobs, and 2 workers processing). More workers can even hurt performance, but this is in a local machine, so not very reliable. Anyway here is the code:
```js
'use strict';

var Queue = require('bull');
var queue = new Queue('test');
var Promise = require('bluebird');

var cluster = require('cluster');
var numWorkers = 2;

module.exports = function (options, cb) {

  if(cluster.isMaster){
    for (var i = 0; i < numWorkers; i++) {
      cluster.fork();
    }

    cluster.on('online', function(worker) {
      console.log('one worker online', worker.process.pid);
      if(--numWorkers == 0){
        var finished = 0;
        var finishTime, startTime;

        var reportResult = function (result) {
          finished += 1;
          if (finished === options.numRuns) {
            finishTime = (new Date()).getTime();
            cb(null, finishTime - startTime);
          }
        };

        // Lets create a few jobs for every created worker
        var prevJobId = 0;
        queue.on('global:completed', function(jobId){
          reportResult();
        })

        queue.on('registered:global:completed', function(){
          queue.isReady().then(function(){
            var start = startTime = Date.now();
            var ops = [];
            for (var i = 0; i < options.numRuns; i++) {
              ops.push(queue.add({i: i}));
            }

            Promise.all(ops).then(function(){
              console.log("Added jobs time:", Date.now() - start);
              //cb(null);
            });
          });
        })
      }
    });

    cluster.on('exit', function(worker, code, signal) {
      console.log('worker ' + worker.process.pid + ' died');
    });
  } else{
    console.log('setting processor')
    queue.process(options.concurrency, function (job, done) {
      return done();
    });
  }
};

```

bradvogel:
I haven't run that test, but it sounds interesting. Do you think the three queues would perform differently from each other in a test run with cluster?

manast:
@bradvogel no idea, I guess there is only one way to know :)

bradvogel:
@lewisjellis this is all ready to merge

LewisJEllis:
My one hesitation here was this comment:
> If I'm interpreting these numbers correctly, it looks like bee-queue is significantly faster on Node 7 and even 6 than on 8, while other queues exhibit more consistent performance numbers across Node versions. Any idea why that might be? Is it worth pointing out in the readme? It looks like we get a 50% throughput boost just by running on 7 instead of 8.
  
I've opened #79 tracking that curiosity - it doesn't need to block merging this otherwise good-looking PR. I rebased and this should be good to merge assuming CI passes.

PR #76, opened 2017-08-24
Document the "stalled" event.

bradvogel:


bradvogel:
@LewisJEllis can we merge?

README.md
```diff
@@ -355,6 +355,16 @@ queue.on('failed', (job, err) => {
...
 
+#### stalled
+
+```js
+queue.on('stalled', (jobId) => {
+  console.log(`Job ${jobId} stalled and will be reprocessed`);
+});
+```
+
+This queue detected that a job [stalled](#stalling-jobs). Note that this might not be the same queue that processed the job and ultimately stalled; instead, it's the queue that happened to _detect_ the stalled job.
```
LewisJEllis: might be slightly more clear to specify "queue instance" rather than just "queue" here - I think of the "queue" as the construct represented by Redis state, and there may be 0 or more active bee-queue instances of the queue. This otherwise this might give impression that if I have two separate queues A and B, an instance of B might detect a stalling job in A, which is not the case.

bradvogel: Yes - you're right. Fixed.

PR #80, opened 2017-09-26
Typescript definitions

pbadenski:


bradvogel:
Hi, thanks for the PR. I'm not deeply familiar with Typescript  - can you share what this file is for?

pbadenski:
Sure thing! :)

Not sure how much you've heard about TypeScript, so for the sake of completeness: "TypeScript is a typed superset of JavaScript that compiles to plain JavaScript" (http://www.typescriptlang.org/).

If a library (ie. **bee-queue**) is written in plain Java Script type information is not available. You can provide an additional file (**index.d.ts**) with all type definitions (classes, function, methods).

After having **index.d.ts** included in **bee-queue** we can use it with type information. This enables type verification, content assist, etc.

LewisJEllis:
Thanks for the contribution! This generally looks good as far as my familiarity goes, but I have one concern/question from past experience with other libraries that ran into maintenance overhead with keeping the typescript definitions updated. I think bee-queue's API is relatively stable and so I'm not quite as concerned, but I'm curious if you have thoughts on anything we could do to help with that -
 automatically testing the typescript definitions in some way, checking if they're out of date, maybe version them to npm package versions, or anything like that to provide visibility/help remind maintainers to update them? I am not familiar enough to know what common practices are around this.

pbadenski:
Hmm I completely understand the concern. I don't have much experience with that. I will research the topic and get back to you.

pbadenski:
Based on what I found, there doesn't seem a good way to handle it.. There's an alternative route, where I could publish the typings to the independent/separate repository (https://github.com/DefinitelyTyped/DefinitelyTyped). It doesn't really make the problem of compatibility go away, but at least gives the user some wiggle room (if there's a type issue you can for instance try to combine different versions).

LewisJEllis:
My understanding is that it's preferable to have types in the project itself if possible and DefinitelyTyped is a fallback. I've actually been starting to use TypeScript lately and learned a bit as well, at least enough that I'd be able comfortable updating the type definitions for minor changes, so I'll go ahead and merge this. I'm not able to immediately publish new version with them, but will do so soon.

notedit:
when we publish this to  https://github.com/DefinitelyTyped/DefinitelyTyped?

mscharley:
@LewisJEllis I just tried to pull bee-queue into a new TypeScript project based on it being in master but it looks like there still isn't a new version with the typings published. When do you plan on publishing a new version?

LewisJEllis:
Merged #89 and published 1.2.0 with `index.d.ts` included.

PR #85, opened 2017-10-31
fix Bluebird promise warning spam

hhhonzik:
with the latest release I got a loads of warnings like this: 
```js
(node:20972) Warning: .then() only accepts functions but was passed: [object Undefined], [object Null]
```

Few minutes diving into this, it looks like bluebirdjs is printing warnings when you pass `null` in a chain `Promise.reject().catch(null)`.


skeggse:
Thanks for this! I updated the PR to pass in an no-op function in the case that there's no callback so it won't trigger an unhandled rejection that crashes the process.

skeggse:
Released in `v1.2.3`.

PR #87, opened 2017-12-20
Fixing quickstart guide in README

mvhirsch:
Hi,

I just tried out the quickstart example and wasn't able to get it working. 
After a quick look, I found two bugs in there :-)

~ greets


LewisJEllis:
Oh nice, thanks for the fix! Sorry I missed this - looks like you fixed #99 ahead of time and then some :)

PR #89, opened 2018-01-06
Typescript definition fix

martinwepner:
This pull request fixes two issues in the index.d.ts for the Typescript definitions.

1. The 'succeeded' event passed an error instead of a result. If the handler function ends with an error, the 'retrying'/'error' event will be called and not the 'succeeded'. 
The fix just changed the type from _Error_ to _any_.

2. The 'options' parameter in the BeeQueue constructor takes optionally a [Redis-options object](https://github.com/NodeRedis/node_redis#rediscreateclient). In the original implementation, you were very limited adding the Redis options. E.g. it was not possible to provide a Redis password which is essential in many cases. 
The fix imports the original Redis options type _ClientOpts_ from the Redis module and uses it as type for the constructor parameter 'options.redis'

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,3 +1,5 @@
+import { ClientOpts } from 'redis';
+
 declare class BeeQueue {
   name: string;
   keyPrefix: string;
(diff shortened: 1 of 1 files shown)


martinwepner:
The check maybe fails because node v6 does not support the _import_ statement. If we use _require_ instead of _import_, we will lose IntelliSense - which would be no option as well in my opinion. 

Not sure what to do here. Maybe someone knows how to work around the problem + maintaining IntelliSense.

LewisJEllis:
Thanks for the fix! I'm not sure what the CI issue was, but retrying the build a couple times fixed it.

LewisJEllis:
Published v1.2.0 to npm with `index.d.ts` included.

PR #98, opened 2018-01-20
fix: DoneCallback.result is optional

brickyang:
e.g. `done(error)`


LewisJEllis:
Thanks! Publishing 1.2.2.

Issue #107, opened 2018-03-28
Custom progress events


stilliard:
Hey, Is there a way to add custom progress pub/sub style events currently? 
e.g. during processing the queue jobs i have a few different types of events to let any subscribers hook into. 
I could implement that with node-redis-pubsub or similar though if not. 

Alternatively, I could look into adding such functionality in? 
Maybe a job could send these with `job.publish('event-name', { data })` and listeners could use the existing `job.on('event-name', console.log)`. 
What do you think?

stilliard:
Closing in favour of adding custom data to progress events. 

Though in case anyone else needs custom event pubsub functionality, there's a [fork](https://github.com/stilliard/bee-queue/tree/custom-job-events) for that available that can be installed with:
"bee-queue": "stilliard/bee-queue#custom-job-events"

Issue #39, opened 2016-02-12
Pass optional data in progress


mpinger:
As a new feature, it would be helpful to be able to pass an optional object or message in the progress() call. The way it is done in kue would be fine.

Thanks,
mike

LewisJEllis:
Sounds reasonable, I'll add this to the upcoming list.

Rush:
I've got the same request.

jeremyblalock:
It's relatively easy to remove the requirement for `progress` to be a number from 0–100. Then you could just pass an object with whatever values you want. Is there any downside to this approach? (I have tested it **[here](https://github.com/jeremyblalock/bee-queue)**, and it seems to work fine)

stilliard:
I have a similar use case, but to pass data as different types of progress events: #107

LewisJEllis:
@Rush @jeremyblalock I wrote some thoughts in #108 on some of my reservations toward this; if possible, I would like to get your feedback on your intended/imagined use cases for this so I can understand them and make sure we support them well. I especially want to make sure that those use cases are actually best served by progress events with arbitrary data (despite their caveats), rather than possibly by the usage of additional queues.

jeremyblalock:
@LewisJEllis in my case I need that flexibility for two reasons:

1. To display both numerator & denominator in the progress (ex: **"Processing Page  5 / 11"**)
2. To display information about the current stage of the job.
    - In my case, users upload a multi-page file to a webserver, which then creates a processing job in bee-queue. The document is then processed page-by-page, and I want to show the title of the current page.

I was able to accomplish all this with my fork, by simply commenting out the code that does type checking on the progress value: https://github.com/jeremyblalock/bee-queue/commit/3fce87b3b71183ca8b8a752dceda272681daad44

LewisJEllis:
Great, thanks for the rundown and specific examples! Those definitely sound like use cases where the context of "where was the event created" matters, the data is only available after starting the job processing, and we need to make sure the data makes it back specifically to whomever created the job. It's similar in spirit to the "progress bar for encoding a video upload" use case of the numeric progress.

PR #108, opened 2018-03-28
Custom job events with job.publish(name, data)

stilliard:
Ref #107.

Created a job.publish() method instead of .emit as originally proposed as the job class extends EventEmiiter and so already had an emit method. *(happy to rename the method to whatever fits best though :+1:)*

README updated & test added to maintain coverage.

One use case for this would be if you were performing a long task such as reading lines in a large file you could publish as you found specific strings you were hunting for. Or if you were running a crawl via this you could then publish events about other urls discovered etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -741,6 +741,30 @@ queue.process(async (job, done) => {
 
 Reports job progress when called within a handler function. Causes a `progress` event to be emitted. Does not persist the progress to Redis, but will store it on `job.progress`, and if other `Queue`s have `storeJobs` and `getEvents` enabled, then the `progress` will end up on all corresponding job instances.
 
+### Job#publish(name,data)
+
+```js
+const job = queue.createJob({...}).save();
+
+job.on('found-something', (data) => {
+  console.log(`Job ${job.id} updated with:`, data);
+});
+
+queue.process(async (job) => {
+
+  // do some work
+
+  job.publish('found-something', {msg:"found abc"});
+
+  // do some work
+
+  job.publish('found-something', {msg:"found xyz"});
+
+});
+```
+
+While processing a job you can send custom events back e.g. for specific log information etc.
+
 #### Job#remove([cb])
 
 ```js
--- a/lib/job.js
+++ b/lib/job.js
@@ -161,6 +161,24 @@ class Job extends Emitter {
     return promise;
   }
 
+  publish(name, data) {
+    // publish a custom event via the pubsub
+
+    const promise = this.queue._commandable().then((client) => {
+      const publishPromise = helpers.deferred();
+      const payload = JSON.stringify({
+        id: this.id,
+        event: name,
+        data
+      });
+      client.publish(this.queue.toKey('events'), payload,
+        publishPromise.defer());
+      return publishPromise;
+    });
+
+    return promise;
+  }
+
   remove(cb) {
     const promise = this.queue.removeJob(this.id).then(() => this);
     if (cb) helpers.asCallback(promise, cb);
--- a/test/job-test.js
+++ b/test/job-test.js
@@ -131,6 +131,22 @@ describe('Job', (it) => {
     });
   });
 
+  it.describe('Custom event', (it) => {
+    it.cb('let you emit & subscribe, passing data', (t) => {
+      const {makeJob} = t.context;
+
+      makeJob().then((job) => {
+        job.on('custom-event-1', (data) => {
+          if (! data || data.hello !== 'world') {
+            throw new Error('failed to recieve correct data');
+          }
+          t.end();
+        });
+        job.publish('custom-event-1', {hello: 'world'});
+      });
+    });
+  });
+
   it.describe('Remove', (it) => {
     it('removes the job from redis', async (t) => {
       const {queue, makeJob} = t.context;


LewisJEllis:
Hmm, interesting, thanks for the PR! 

I'd been meaning to expand on some of these thoughts in #39 (and relatedly #107), but since we have a proposal here, I'll do it here.

My main immediate implementation-specific concern for this PR is that we need to avoid event namespace collision; we already publish/emit some [job events](https://github.com/bee-queue/bee-queue#job-events) as part of the job's lifecycle, and user-published events should not be able to get mixed in with these.

We could force some unique prefix for all event names, but I also wonder if the event name actually even matters. If we allowed `progress` events to report arbitrary data as suggested in #39, would that satisfy your desired use case? It seems like it would, based on examples here, but let me know if you think not.

This leads to my more skeptical "is this actually necessary" concern that would also apply to allowing `progress` events to send arbitrary data. I have generally been hesitant to consider expanding on pubsub-based events/communication for a few reasons:
- it all operates in few-guarantees-available redis pub-sub land: best-effort only, hard to ensure delivery in the face of failures; don't want users to become dependent on things we can't guarantee
- a lot of such use cases can also be supported by just having another queue that the worker pushes jobs into (queue A's worker function can create new jobs for queues B and C, etc)
  - I'm curious if your use case(s) could be satisfied by this pattern, or if it's specifically important that the *producer* of the job finds out about these events/updates
- crosstalk overhead: all job events are sent to every queue that is listening for queue pubsub events; we have no way to send a job event just to the producer queue of that job, unless that producer queue is the only queue instance for that queue name with `getEvents` enabled
  - we could write about this in the readme, and change config in v2.0 to be more off-by-default, but this is still feels sort of bleh, and I don't think we should start trying to keep track of which producer enqueued a job

For context, the reason I introduced the job pub/sub events at all was for use cases where the creator of the job is specifically waiting for the job to *finish* before doing something else (like responding to the HTTP request that caused the job to be enqueued). The numeric `progress` event as it exists currently was really just added as a sort of "well, kue has it, might as well just for kicks", but I've never really used it myself. It seems like if you are interested in partial-progress reports, you might be better served by breaking things down into multiple jobs, each with their own completion and finishing guarantees (which pubsub events do not have). The arbitrary event notification capability feels to me like it should only be useful/needed if the context of *where* the job was created is considered important (and I suspect that is usually not the case). Of course, there can definitely be such use cases - I imagine there are, but before trying to support them I want to understand what they are specifically, and consider/understand why additional jobs/queues won't satisfy them as well.

/cc @bradvogel curious for your thoughts - do you have use cases where you would find the ability to pubsub arbitrary progress data useful, even given the caveats/lack-of-guarantees discussed above, or have you moreso seen "queue A's worker might create jobs on queues B and C" type uses?

stilliard:
Hi @LewisJEllis , thanks for getting back to me on this. 
I fully understand your concerns & I would also not want people to become reliant on events that have no guarantee of being delivered. In my case that's not an issue though I should have provided a better example. 

My current use case for this is an application i'm working with that uses websockets to update a sort of log of information back to the user based on different things found during a crawl. 
E.g. specific content types discovered, issues etc. & so for me having the different event names makes sense to separate these out. 
The progress could also be used here and i could create an object that then had a type property to let me know which type of data it is, but seems just as simple to have a way to break these up. Also I have no need to hold these as the "progress" as such & didn't want to break any existing requirements on progress being a number between 0 and 100. 
I do separately use progress, but these events are separate and they themselves don't actually affect the progress, they are more more additional information I can send back to the user to let them know what's going on. 
The data in my case the important data from these events is actually also being saved to a database anyway so the use case for this for me is to avoid polling the database and get the events live as they happen back from the job, and be able to then send these back to the user via websockets.

Hope that helps clarify why for my use case having a way a separate method outside of progress is useful & that there's no need for 100% deliverability. 

Ref the namespace collision issue, I can make a change to give these events a prefix so that they will not cause any issues with the existing events?

We can also name this method however you like to make it clear that it's for non essential stuff and would not be persisted?

LewisJEllis:
> The progress could also be used here and i could create an object that then had a type property to let me know which type of data it is

Yes, I was thinking of the same workaround; I think this would also be preferable for the sake of having to register just a single event listener and having a constant fixed set of possible events that a `job` instance will emit.

> the important data from these events is actually also being saved to a database anyway from inside the job handler

Got it, hence we can be less concerned about delivery guarantees - if any one of the three players (redis, worker, web/socket server) goes down, the live realtime session would have to be aborted anyway, and the results of the crawl would be viewed later in another request that reads from the database.

> so the use case for this for me is to avoid polling the database and get the events live as they happen back from the job, and be able to then send these back to the user via websockets.

Makes sense - this is aligned with my original motivation for job events (to return the results of the job within the HTTP request that caused it to be created), just for a more real-time incremental situation.

Okay, I think between your explanation/example use case and the use cases given in #39, I am satisfied that this mechanism in particular has legitimate usefulness and that the weak guarantees aren't consequential to those uses. Thank you for helping to convince me.

For the implementation, I'd prefer to go with the "just allow `progress` events to have arbitrary data" approach like @jeremyblalock implemented [here](https://github.com/jeremyblalock/bee-queue/commit/3fce87b3b71183ca8b8a752dceda272681daad44). If you want to update this PR for that approach, I'll be happy to review, but if not I can take care of cherry-picking Jeremy's implementation into a branch and updating the tests, readme, etc from there.

stilliard:
No worries. For my project atm it makes more sense to use named events as rather than creating an abstraction on top of the jobs to filter between different types & to keep progress as a separate numerical value, but that's alright as I can keep using my forked version. 
Good luck with the other implementation, I'll close this pull request now :+1:

PR #123, opened 2018-06-04
Fix "too many values to unpack" while raising delayed jobs to waiting state

mattbornski:
{ ReplyError: ERR Error running script (call to f_7c74c5918ff75c88a0db8507b3c0a4ddf34845b3): @user_script:17: user_script:17: too many results to unpack 
at parseError (/var/task/node_modules/redis-parser/lib/parser.js:193:12)
at parseType (/var/task/node_modules/redis-parser/lib/parser.js:303:14)
command: 'EVALSHA',
args: 
[ '7c74c5918ff75c88a0db8507b3c0a4ddf34845b3',
2,
'bq:queue_name:delayed',
'bq:queue_name:waiting',
1528136012927,
1000 ],
code: 'ERR' }

### DIFF
--- a/lib/lua/raiseDelayedJobs.lua
+++ b/lib/lua/raiseDelayedJobs.lua
@@ -10,13 +10,21 @@ returns number of jobs raised and the timestamp of the next job (within the near
 local now = tonumber(ARGV[1])
 
 -- raise any delayed jobs that are now valid by moving from delayed to waiting
-local raising = redis.call("zrangebyscore", KEYS[1], 0, ARGV[1])
-local numRaising = #raising
+local offset = 0
+-- There is a default 1024 element restriction
+local count = 1000
+while true do
+  local raising = redis.call("zrangebyscore", KEYS[1], 0, ARGV[1], 'LIMIT', offset, count)
+  local numRaising = #raising
+  offset = offset + numRaising
 
-if numRaising > 0 then
-  redis.call("lpush", KEYS[2], unpack(raising))
-  redis.call("zremrangebyscore", KEYS[1], 0, ARGV[1])
+  if numRaising > 0 then
+    redis.call("lpush", KEYS[2], unpack(raising))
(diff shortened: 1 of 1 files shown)


lib/lua/raiseDelayedJobs.lua
```diff
@@ -10,13 +10,21 @@ returns number of jobs raised and the timestamp of the next job (within the near
 local now = tonumber(ARGV[1])
 
 -- raise any delayed jobs that are now valid by moving from delayed to waiting
-local raising = redis.call("zrangebyscore", KEYS[1], 0, ARGV[1])
-local numRaising = #raising
+local offset = 0
+-- There is a default 1024 element restriction
```
skeggse: is this a redis `lpush` limitation or a lua `unpack` limitation? can you comment here to clarify?

skeggse: nit: `s/There/there`

mattbornski: I believed this limit is derived from the Lua config parameter, DEFAULT_STACK_SIZE (http://www.lua.org/source/4.0.1/llimits.h.html#DEFAULT_STACK_SIZE); however, upon deeper inspection, Redis uses Lua 5.1 and I can't find an equivalent parameter there.  It is possible this 1024 assertion is spurious; 
the closest equivalent I can find in Lua 5.1 is LUAI_MAXCSTACK and it's set to 8000 in Redis (http://download.redis.io/redis-stable/deps/lua/src/luaconf.h)

mattbornski:
@skeggse I haven't been using bee-queue for quite a while; I would point out that for people who encounter the error I did, either the PR as-is or your proposed change would be a dramatic improvement, because *no* jobs were being processed due to the error once a critical size was reached.

skeggse:
Closing in favor of #192.

PR #128, opened 2018-06-26
Added NOTE to set activateDelayedJobs of a Queue to be able to use the Job#backoff functionality

hellojoshuatonga:
Just a little documentation improvement. I spent hour(s) to make the Job#backoff work on my setup. Turned out I just missed a little detail.  I even created a hackish way for the retry and backoff strategy

PR #130, opened 2018-07-05
Update type definitions

bondz:
I've made minimal changes to the type definitions to make it easier to consistently validate the data type that a queue or job is processing.

cc @pbadenski @martinwepner

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -25,6 +25,9 @@ declare class BeeQueue<T = any> extends EventEmitter {
   on(ev: "job failed",    fn: (jobId: string, err: Error) => void): this;
   on(ev: "job progress",  fn: (jobId: string, progress: number) => void): this;
 
+  ready(): Promise<this>;
+  ready(cb: (err: Error | null) => void): this;
+
   createJob<U>(data: U): BeeQueue.Job<T>;
 
   getJob(jobId: string, cb: (job: BeeQueue.Job<T>) => void): void;
--- a/index.d.ts
+++ b/index.d.ts
@@ -28,7 +28,7 @@ declare class BeeQueue<T = any> extends EventEmitter {
   ready(): Promise<this>;
   ready(cb: (err: Error | null) => void): this;
 
-  createJob<U>(data: U): BeeQueue.Job<T>;
+  createJob<U extends T>(data: U): BeeQueue.Job<U>;
 
   getJob(jobId: string, cb: (job: BeeQueue.Job<T>) => void): void;
   getJob(jobId: string): Promise<BeeQueue.Job<T>>;
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,6 +1,10 @@
+/// <reference types="node" />
+/// <reference types="redis" />
+
+import { EventEmitter } from 'events';
 import { ClientOpts } from 'redis';
 
-declare class BeeQueue {
+declare class BeeQueue<T = any> extends EventEmitter {
   name: string;
   keyPrefix: string;
   jobs: any;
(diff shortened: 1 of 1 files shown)


bondz:
The test failure is unrelated to this commit.

hellojoshuatonga:
> Looks like the new version of eslint isn't compatible with Node 4, and we run travis against Node 4. - @skeggse #128

hellojoshuatonga:
The `eslint` version is already pinned in #128 
Just rebase after it got merged

index.d.ts
```diff
@@ -11,28 +15,31 @@ declare class BeeQueue {
...
   on(ev: "stalled",   fn: (jobId: string) => void): this;
 
   on(ev: "job succeeded", fn: (jobId: string, result: any) => void): this;
   on(ev: "job retrying",  fn: (jobId: string, err: Error) => void): this;
   on(ev: "job failed",    fn: (jobId: string, err: Error) => void): this;
   on(ev: "job progress",  fn: (jobId: string, progress: number) => void): this;
 
-  createJob<T>(data: T): BeeQueue.Job;
+  ready(): Promise<this>;
+  ready(cb: (err: Error | null) => void): this;
```
calebboyd: The callback here is optional and Promise<this> is always returned.

bondz: Yes, you're right @calebboyd Fixed that.

skeggse:
This has languished for a long time - is it still worth looking at? What does the introduction of the `T` parameter do to the module's interface? Would this be a breaking change?

skeggse:
cc @hughsw - I think you're more experienced with typescript than I am.

bondz:
Yes, it could be a breaking change for some typescript users, although adding the generic doesn't change anything, only letting users be able to tell the type of data a job is processing.

skeggse:
Would it force typescript users of the library to update their code to add explicit annotations?

bondz:
No. Inference would be better and it should only catch errors.

skeggse:
Ok cool - that sounds good to me!

index.d.ts
```diff
@@ -1,6 +1,10 @@
+/// <reference types="node" />
+/// <reference types="redis" />
+
+import { EventEmitter } from 'events';
 import { ClientOpts } from 'redis';
 
-declare class BeeQueue {
+declare class BeeQueue<T> extends EventEmitter {
```
martinwepner, line 7: @bondz sorry, for some reason I missed your PR and my mention.

I have briefly skimmed your PR and I really appreciate it! However I would suggest two changes: 

```suggestion
declare class BeeQueue<T = unknown> extends EventEmitter {
```

This will fix the following issue:
```typescript
import bq from "bee-queue"
class Foo {
	private _queue: bq; // This will result in error "Generic type 'BeeQueue<T>' requires 1 type argument(s)."
}
```

bondz: Yes, I see what you mean. Don't you think `any` would be easier for users, same as being untyped. `unknown` is best, I know but users would have to explicitly cast or validate the data type everywhere.

index.d.ts
```diff
@@ -11,28 +15,33 @@ declare class BeeQueue {
...
 
-  createJob<T>(data: T): BeeQueue.Job;
+  ready(): Promise<this>;
+  ready(cb?: (err: Error | null) => void): Promise<this>;
+
+  isRunning(): boolean;
 
-  getJob(jobId: string, cb: (job: BeeQueue.Job) => void): void;
-  getJob(jobId: string): Promise<BeeQueue.Job>;
+  createJob(data: T): BeeQueue.Job<T>;
```
martinwepner, line 33: second change I would suggest (not very beautiful but backwards compatible):
```suggestion
  createJob<U extends T>(data: U): BeeQueue.Job<U>;
```

This will fix the following: 
```typescript
let job = this._queue.createJob<PayloadType>(payload) // Expected 0 type arguments, but got 1
```

index.d.ts
martinwepner: Yes you are correct! I did not mention this issue, because I casted the type to the correct one😅
any would be better for backwards compatibility. I will edit my suggestion

index.d.ts
bondz: Would this still be a problem if T defaults to `any`?

martinwepner: I don't think so. In the old version one generic may have been passed. Now no generic is allowed, so all old users how used the generic will get this tsc error:
// Expected 0 type arguments, but got 1

bondz: Right. Thanks. Changes applied.

skeggse:
Let me know when this is good to go!

bondz:
@skeggse This is fine now :) Right @martinwepner ?

martinwepner:
Right! From my side there is nothing more to complain about.👍🏽

PR #134, opened 2018-07-13
feat: support custom strategies on a queue

calebboyd:
Enables providing a custom function to determine an appropriate delay.
Must be set on all queue instances.

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -20,6 +20,7 @@ class Queue extends Emitter {
     this.jobs = new Map();
     this.activeJobs = new Set();
     this.checkTimer = null;
+    this.backoffStrategies = new Map(backoff);
 
     this._closed = null;
     this._isClosed = false;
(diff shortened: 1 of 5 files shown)


lib/queue.js
```diff
@@ -526,7 +527,7 @@ class Queue extends Emitter {
         ? job.options.backoff.strategy
         : 'immediate';
       const strategy = job.options.retries > 0
-        ? backoff.get(strategyName)
+        ? this.backoffStrategies.get(strategyName)
```
skeggse, line 530: So the behavior here if the backoff strategy hasn't been configured (maybe due to deployment issues in multi-service scenarios) is that the job would get retried immediately? That may be a suboptimal result - not sure. It also seems like we should report this as an error for operational visibility. Not sure if that should be an `error` event or something else. Thoughts?

skeggse: tbh can't remember what I was thinking. seems like this matches current behavior, albeit with a little more "buyer beware."

skeggse (approved):

PR #154, opened 2018-12-12
fix(backoff): allow no delay arg when setting immediate strategy

ghmeier:
Per [the documentation](https://github.com/bee-queue/bee-queue#jobbackoffstrategy-delayfactor) you should be able to call `job.backoff('immediate')` to set an immediate backoff strategy for retried jobs. However, the `backoff` method explicitly disallowed this by checking for `Number.isSafeInteger(undefined)` which returns false. This means you currently cannot specify an immediate backoff strategy without also specifying a positive delay value.

To fix this, only check for valid integers when the strategy is not `immediate`.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -85,7 +85,7 @@ declare namespace BeeQueue {
 
     setId(id: string): this;
     retries(n: number): this;
-    backoff(strategy: "immediate" | "fixed" | "exponential", delayFactor: number): this;
+    backoff(strategy: "immediate" | "fixed" | "exponential", delayFactor?: number): this;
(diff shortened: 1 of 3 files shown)


ghmeier:
Bump on this ☝️

bradvogel:
@ghmeier I just added you to the Bee Queue org, so you can merge this.

@LewisJEllis - Garret is an engineer on our team

hughsw:
The `delayFactor` arg to `backoff()` in `index.td.s` should be made optional:
```
  backoff(strategy: "immediate" | "fixed" | "exponential", delayFactor?: number): this;
```

Issue #95, opened 2018-01-14
calling done with error doesnt fail the job on Arena


sharonass:
I've setup a  bee-queue with arena and In my local machine when I call 

queue.process(async function (job, done) {

        console.log(`Processing job ${job.id}`);
        job.reportProgress(5);
        let res = await my_service.execute(job);
        if(res.status) {
          //report job progress
          job.reportProgress(100);
          return done(null, res.msg);
        } else {
          done(new Error(res.msg));
          
        }
      });    

I see on my flow that the done(new Error(res.msg)); is being called and I also debugged the done function and I see it rejects the promise. however when I look at the job on Arena I see it in success section

sharonass:
I also dont get fail event

queue.on('failed', (job, err) => {
    console.log(`Job ${job.id} failed with error ${err.message}`);
  });

dylanjha:
@sharonass I think the problem is that if you pass an `async` function (or a promise) the `done` callback will be useless

See docs:

> N.B. If the handler returns a Promise, calls to the done callback will be ignored.

https://github.com/bee-queue/bee-queue#queueprocessconcurrency-handlerjob-done

LewisJEllis:
Aha, I was a bit puzzled here but @dylanjha I think you hit the nail on the head.

I would replace:
```js
return done(null, res.msg);
...
done(new Error(res.msg));
```
with:
```js
return res.msg;
...
throw new Error(res.msg);
```

PR #155, opened 2018-12-13
add note to README about done() and process passed async function

cymen:
I found this aspect of how bee-queue works confusing. It makes sense after reading #95 however it would be great to call it out in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -342,7 +342,7 @@ queue.on('failed', (job, err) => {
 });
 ```
 
-This queue has processed `job`, but its handler reported a failure either by rejecting its returned Promise, or by calling `done(err)`.
+This queue has processed `job`, but its handler reported a failure either by rejecting its returned Promise, or by calling `done(err)`. Note that if you pass an async function to process, you must reject it by returning `Promise.reject(...)` or throwing an exception ([done does not apply](https://github.com/bee-queue/bee-queue/issues/95)).
 
 #### stalled
 


cymen:
I realized done is mentioned in a couple of places. It might be better to add a new section with an explanation and linking to that with a short summary of `done(err)` vs `Promise.reject`/`throw`. That section might include some example code. Perhaps something like this:

worker.js
```js
import Queue from 'bee-queue';

const queue = new Queue('test-retry', {
  isWorker: true,
  removeOnSuccess: true,
});

const key = 'count';
let count = 0;

async function run() {
  count += 1;
  if (count >= 5) {
    count = 0;
    return true;
  }
  return false;
}

queue.on('ready', () => {
  queue.process(async job => {
    const result = await run();
    if (!result) {
      return Promise.reject(new Error('my custom error'));
    } else {
      return 'finished';
    }
  });
});
```

producer.js
```javascript
import Queue from 'bee-queue';

const queue = new Queue('test-retry', {
  isWorker: false,
});

async function run() {
  const job = await queue
    .createJob({})
    .retries(5)
    .save();

  job
    .on('succeeded', result => {
      console.log('succeeded!', result);
    })
    .on('retrying', error => {
      console.log('retrying', error.message);
    })
    .on('failed', error => {
      console.log('failed', error);
    })
    .on('progress', progress => {
      console.log('progress', progress);
    });
}

run();
```

PR #169, opened 2019-06-27
feat: recurring jobs

xphoniex:
This feature was requested frequently so I decided to add it.

I'm prefixing my key with `r:[ms]:` and inside `raiseDelayedJobs.lua`, pre-emptively schedule such keys for next run. 

I keep a `.count` for number of times this has been run as well, which could come in handy.

To stop the job you'd simply call remove.


giorgiofellipe:
This would be great! Any news?

compwright:
Closing due to no response.

PR #162, opened 2019-03-15
If the brpoplpush commands fails, wait before retrying

tmeasday:
The current behaviour is to retry as fast as possible, leading to a spam of commands and error messages if the redis instance is for example unavailable for a brief period. It seems more sensible to take a bit more care in retrying.

PR #175, opened 2019-10-18
Delay failed jobs on handled worker

Lucaber:
Add queue setting to prevent a failed job to be scheduled on the same
worker for `delayFailedJobsByWorker` ms.
Disabled by default.


compwright:
This is a duplicate of #162, closing

PR #178, opened 2020-01-27
fix(queue): always close eclient conn if it exists

ethanresnick:
Before, the eclient was created if this was true (see line 77):

```
this.settings.getEvents || this.settings.activateDelayedJobs
```

But then it was closed iff `this.settings.getEvents` was true, which
meant that `close()` would fail to terminate all connections if the
settings were `{ getEvents: false, activateDelayedJobs: true }`.

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -195,7 +195,7 @@ class Queue extends Emitter {
       if (this.settings.quitCommandClient) {
         clients.push(this.client);
       }
-      if (this.settings.getEvents) {
+      if (this.eclient) {
         clients.push(this.eclient);
       }
 


ethanresnick:
@skeggse FYI, the travis failure seems unrelated to this change; looks like eslint ships some modern JS code that's not compatible with older versions of Node

skeggse:
@ethanresnick good catch, thanks!

ethanresnick:
Sure thing. Thanks for merging so quickly :)

Any chance a new version can be published to npm with this fix? I need it for my current project, and would rather not have to create a distinct npm package to depend on

skeggse:
Released in `v1.2.3`.

ethanresnick:
Thanks!

PR #180, opened 2020-01-28
fix(types): account for Queue.close(timeout)

ethanresnick:
These types support all four call signatures:

```
close(): Promise<void>
close(timeout: number): Promise<void>;

close(cb: () => void): void;
close(timeout: number, cb: () => void): void;
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -41,8 +41,9 @@ declare class BeeQueue {
   checkHealth(): Promise<BeeQueue.HealthCheckResult>;
   checkHealth(cb: (counts: BeeQueue.HealthCheckResult) => void): void;
 
-  close(): Promise<void>;
   close(cb: () => void): void;
+  close(timeout?: number): Promise<void>;
+  close(timeout: number, cb: () => void): void;
 
   removeJob(jobId: string): Promise<void>;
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -41,8 +41,9 @@ declare class BeeQueue {
   checkHealth(): Promise<BeeQueue.HealthCheckResult>;
   checkHealth(cb: (counts: BeeQueue.HealthCheckResult) => void): void;
 
-  close(): Promise<void>;
   close(cb: () => void): void;
+  close(timeout?: number): Promise<void>;
+  close(timeout: number, cb: () => void): void;
```
skeggse: The `timeout` is technically optional here, so you can pass something like `close(timeoutOrNull, callback)`. I dunno if this works in typescript, though:
```suggestion
  close(timeout?: number, cb: () => void): void;
```

ethanresnick: TS won't allow an optional parameter (with the `?`) before a required one, but I changed this to:

```ts
close(timeout: number | undefined | null, cb: () => void): void;
```

skeggse: sgtm

skeggse:
Released in `v1.2.3`.

Issue #184, opened 2020-02-20
Queue Local 'failed' called instead of 'retrying' event


hughsw:
I'm evaluating bee-queue for use in a client's project.  I am not able to get the 'retrying' Queue Local Event to fire.  Instead the 'failed' event gets fired, even though the job is actually retried.  So it appears that the job has failed even though it might actually be being retried.

The Queue PubSub Event 'job retrying' does get fired.  But I don't want the overhead of this PubSub event nor the complexity of its job-level configuration for our plugins.

```javascript
// BeeQueue Code demonstrating that Queue Local Events 'retrying' doesn't work
// Also of note: backoff 'immediate' strategy requires a positive delay argument

const Queue = require('bee-queue');
const redis = require('redis');

const redisClient = redis.createClient({ password: process.env.REDIS_PASSWORD });

const producer1 = new Queue('my-queue', { redis: redisClient, isWorker: false });
const createJob = payload => producer1
      .createJob(payload)
      // Bug: positive delay argument is required to explicitly use 'immediate'
      .backoff('immediate', 1)
      .retries(3)
      .save();

// Worker
const queue = new Queue('my-queue', { redis: redisClient, activateDelayedJobs: true });

// Queue Local Events
//   from https://github.com/bee-queue/bee-queue#queue-local-events
queue.on('succeeded', (job, result) => {
  console.log(`QUEUE succeeded: Job ${job.id} succeeded with result: ${result}`);
});
// Bug: this event doesn't happen, even though the job is retried and the 'job retrying' event below does happen
queue.on('retrying', (job, err) => {
  console.log(`QUEUE retrying: Job ${job.id} failed with error ${err.message} but is being retried!`);
});
queue.on('failed', (job, err) => {
  console.log(`QUEUE failed: Job ${job.id} failed with error ${err.message}`);
});

// Queue PubSub Events
//   from https://github.com/bee-queue/bee-queue#queue-pubsub-events
queue.on('job succeeded', (jobId, result) => {
  console.log(`JOB succeeded: Job ${jobId} succeeded with result: ${result}`);
});
queue.on('job retrying', (jobId, err) => {
  console.log(`JOB retrying: Job ${jobId} failed with error ${err.message} but is being retried!`);
});
queue.on('job failed', (jobId, err) => {
  console.log(`JOB failed: Job ${jobId} failed with error ${err.message}`);
});

// checkHealth
queue
  .ready()
  .then(async queue => {
[...]

PR #185, opened 2020-02-21
fix(backoff): no delay arg when setting immediate strategy, add test

hughsw:
This fixes and tests a minor problem mentioned in #184 

This won't break code that has worked around the problem by providing a delay argument to the immediate strategy.


skeggse:
Actually, gonna favor #154 over this.

PR #186, opened 2020-02-21
fix(queue): emit retrying Queue Local Event instead of failed

hughsw:
This fixes the problem reported in #184 

This fix could conceivably break code that has worked around the problem, e.g. by looking at `job.options.retries` or `job.status` in their queue failed handler...

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -569,7 +569,7 @@ class Queue extends Emitter {
     const promise = helpers.deferred();
     multi.exec(promise.defer());
 
-    return promise.then(() => [status, result]);
+    return promise.then(() => [job.status, result]);
   }
 
   process(concurrency, handler) {
@@ -643,7 +643,7 @@ class Queue extends Emitter {
 
           /* istanbul ignore else */
           if (results) {
-            const status = results[0], result = results[1];
+            const [status, result] = results;
             this.emit(status, job, result);
           }
         });
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -643,7 +643,7 @@ class Queue extends Emitter {
 
           /* istanbul ignore else */
           if (results) {
-            const [status, result] = results;
+            const status = results[0], result = results[1];
             this.emit(status, job, result);
           }
         });
--- a/test/queue-test.js
+++ b/test/queue-test.js
@@ -640,6 +640,25 @@ describe('Queue', (it) => {
       t.is(counts.failed, 1);
     });
 
+    it('reports a retried job', async (t) => {
+      const queue = t.context.makeQueue();
+
+      queue.process(async (job) => {
+        t.is(job.data.foo, 'bar');
+        throw new Error('failed for retry!');
+      });
+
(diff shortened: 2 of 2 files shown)


hughsw:
Apologies for the commit churn -- It's taking a little while to get my build/test configuration running in a Docker container....

lib/queue.js
```diff
@@ -643,7 +643,7 @@ class Queue extends Emitter {
 
           /* istanbul ignore else */
           if (results) {
-            const status = results[0], result = results[1];
+            const [status, result] = results;
```
skeggse: I don't remember exactly, but I think `bee-queue` might still purport to support Node 4.0.0 - this may be a breaking change as the destructuring assignment isn't supported until 6.0.0.

hughsw:
I have undone the destructuring syntax, and fixed the test (grrr, so I thought...  Tests are failing due to timeout; I still have a lot to learn aboue bee-queue and bee-queue testing...).

The question about emitting both 'failed' and 'retrying' is interesting.  My gut reaction is that the two messages is a more surprising behavior than what the docs imply will happen, and what the names of the events imply; but, it doesn't contradict what the docs say, and the retrying docs even mention that the job being retried has just failed...

However, emitting only one of 'retrying' or 'failed' for the queue when a job fails seems less surprising.  It's what the jobs do.  The retrying messages let you know the job is still viable.  'succeeded' or 'failed' are the final word on the job.  Otherwise you have to actively dig around in implementation-specific details to make a determination about the job's final status...

hughsw:
Indeed, the reason I filed the issue, #184 was that Queue Local is a simple way to manage job lifecycle that didn't have the complexity or overhead of the PubSub or the Job callbacks.  It would be very surprising that Queue Local has different semantics and no simple way to know if a job has really failed yet...

skeggse:
I agree with everything you're saying, and I don't think we're in a position to ship a new major revision of the library. Shipping the change that makes `retrying` fire instead of `failed` represents a major change in the API surface and since we don't know what downstream folks are doing with the library this can break their expectations (even though the API does seem like it should only emit one or the other). Maybe as a compromise we can create a new event on the `Queue` like `failure:final` or `finalFailure` that signals that a job has finished and is not being retried, and in `bee-queue` v2 can remove or deprecate the new event in favor of a fixed `failure` event.

Thanks also for pointing out the documentation/confusion problems here - I think we can help with that some with more documentation, but I don't see an optimal/foolproof solution here.

hughsw:
I agree.

Here's the simplest path forward that I can see:
- we make no code changes
- update the documentation to remove any mention of a Queue Local 'retrying' event
- we update the Queue Local 'failed' documentation to be explicit that:
  - all job failures emit the event, regardless of whether the job will be retried
  - using only Queue Local events, there is no simple, stateless way to determine that a job has failed for good (won't be retried)
- we add a comment or two to the code regarding how things could be changed in order to unify the 'succeeded', 'retrying', 'failed' callbacks across the three classes of events

Issue #189, opened 2020-03-08
Reproducible: bee-queue loses track of jobs


hughsw:
I have seen jobs that have successfully completed, but do not emit either of `succeeded` or `failed`.  The problem seems more likely on jobs that finish quickly.  It also may be influenced by the concurrency setting.

I would expect that any job that has had `job.save()` successfully resolve will eventually emit either `succeeded` or `failed`.  However, we can create and save jobs which never emit either of those job-level events.  We have observed this in Docker containers running under Docker for Mac and under Docker on Ubuntu on Digital Ocean droplets.  

Here's a repo that builds images that reliably demonstrate the problem, although not deterministically.

https://github.com/hughsw/stress-beequeue#run-the-bee-queue-client

This underlying issue may be the same as reported here #78

hughsw:
I'm able to get this failure on a variety of cloud platforms (Docker for Mac, VirtualBox, Digital Ocean) and distributions (Ubuntu, Node, Alpine).  It reliably starts failing after things warm up with a few hundred quick jobs.

hughsw:
I have a fix.  It's based on the analysis here #78 which is spot on.  However the 'crafty' ideas for solving the problem didn't pan out for me -- it's a race condition, and as such has to be faced head on rather than toyed with.

I have a fix.  It's successfully run over 128,000,000 jobs on my dev server overnight with no job failures and without dropping a single job.

It's simpler than I had thought, but it changes how job ids get created.  It keeps job-id-creation in the Queue rather than in Redis.

I am working to get the bee-queue test harness working in a Docker container environment.  I will see if I can add a test that exercises this bug, and that thus will prevent a regression.

I'll put up a WIP/RFC PR once I've teased out a bunch of debugging cruft from the code.

hughsw:
I have not forgotten this issue -- I'm using the fix in our deployments.  But, I haven't had time to create the tests and make the minimal changes for the fix.

bradvogel:
Mind putting a PR with the fix? Even a draft PR without tests would be helpful to us to understand the issue more.

jdelman:
> I have not forgotten this issue -- I'm using the fix in our deployments. But, I haven't had time to create the tests and make the minimal changes for the fix.

Do you have a fork up?

hughsw:
A PR with one approach to the problem is here #193

hughsw:
[...]

Issue #78, opened 2017-09-12
'success' event not fired


henchmun:
Hey there,

Thanks for the great library, although unfortunately I have been having a few issues,

I've been using bull and recently switched to bee queue to make use of the job success event callbacks that mean I can use async.js parallel to set of loads of tasks to be done by workers and wait for them all to complete,

My problem comes when the queue says that it has "completed" everything

```js
{ waiting: 0,
  active: 0,
  succeeded: 1250,
  failed: 0,
  delayed: 0,
  newestJob: 1250 }
```

But by my own logging:
```
Checked stalled jobs 0
Total: 1250 jobs
Outstanding: 10 jobs
```

10 Jobs didn't fire a callback, which then hangs my program which is waiting for the success callback of each job before it moves on to the next step

This is probably a scalability issue of where jobs are coming back to the dispatch too quickly to handle, as when jobs have a setTimeout in them of 100ms the queue doesn't hang. But in that case, I would expect for succeeded to be incorrect when calling checkHealth (although I know this is asking Redis instead of the library what it has seen). 

As all jobs have "succeeded" is there a way for me to then recall the data returned by these outstanding jobs? Or is there a way for me to forcefully retry them again?

For now I'm going to attempt to set a manual timeout to remove and save the job again.

Thanks for the hard work though, cheers for any help anyone may have for fixing this,

Hunchmun:
Just following up that my nasty hack does work:

```js
        let timeout = setTimeout(() => {
            console.log("Retrying:", job.id);
            job.remove((err, job) => {
                if (err) {
                    console.error(err);
                    throw err;
                } else {
                    job.save((err, job) => {
                        if (err) {
                            console.error(err);
                            throw err;
                        }
                    });
                }
            });

        }, this.settings.retryTime * 1000);
```

But anyway to ensure the consistency of the queue would be appreciated

Hunchmun:
Apologies for the spam, but I come back to you with some really meaningful data.

Bee-Queue or one of it's dependencies (maybe redis) doesn't handle arrays very well. One of my tasks returns an array, which would then cause all jobs within that lot to back up depending on how many were within the queue. 

Changing the array to an object meant that less jobs were being reported as successfully done without firing the callback. However, under extreme conditions, jobs would still need to be retried and it only managed to perform well with batches of 2500 jobs when the larger return values/objects were omitted.

LewisJEllis:
[...]

PR #193, opened 2020-03-26
fix(job.id): Use unique hex string for job id to avoid a race condition

hughsw:
These changes fix the problems reported in #189 and originally in #78 .  They do so by managing the job id in the Job object itself rather than in Redis.  Various simplifications emerge.  Note that with these changes, the only purpose of the `bq:name:id` key in Redis is to support the `newestJob` field returned by checkHealth().

A couple of tests fail because they rely on the incrementing numerical-properties of the old job-id creation or on the idea that user-provided job ids are special.  Whether user-provided ids are special  isn't clear to me.  So, I haven't tried to fix the tests yet.

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -6,6 +6,13 @@ function has(object, name) {
   return hasOwn.call(object, name);
 }
 
+// A simple way to get 13 statistically random hex digits, not cryptographic
+const MAX_SAFE_NYBBLE = 13; // low-order 52 bits of double mantissa's 53 bits
+const randomHex13 = () => Math.floor(Math.random() * Number.MAX_SAFE_INTEGER)
+  .toString(16)
+  .padStart(MAX_SAFE_NYBBLE, '0')
+  .slice(-MAX_SAFE_NYBBLE);
+
 const promiseUtils = require('promise-callbacks');
 
 module.exports = {
@@ -16,4 +23,5 @@ module.exports = {
   withTimeout: promiseUtils.withTimeout,
   wrapAsync: promiseUtils.wrapAsync,
   has,
+  randomHex13,
 };
--- a/lib/job.js
+++ b/lib/job.js
@@ -49,35 +49,38 @@ class Job extends Emitter {
   save(cb) {
     const toKey = this.queue.toKey.bind(this.queue);
 
+    if (!this.id) {
+      this.id = helpers.randomHex13();
+    }
+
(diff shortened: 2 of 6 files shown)


hughsw:
I don't understand the test failure involving the stack trace contents.

hughsw:
So this change interferes with the estimation of job-creation-throughput that the documentation for `checkHealth()` discusses, which is indirectly exercised by the remaining id-based test failure.

bradvogel:
Can you share more about how generating the job id from calling Redis' `incr` command leads to Bee Queue losing track of jobs?

hughsw:
Please see https://github.com/bee-queue/bee-queue/issues/189 which reproduces the problem.  

Specifically to your question: Redis does the increment and pushes the job onto the waiting queue atomically.  The worker can pop the job off the waiting queue and get it finished before the job's ID has been set in the queue's container (which happens asynchronously with the worker running the job).  So, when the job finishes, the queue does not emit a succeeded event because the queue doesn't yet know about the job.

bradvogel:
I see now - thanks. Just to confirm - the only symptom of this is that events won't be emitted for those jobs, right?

I'll let @skeggse or @LewisJEllis review.

hughsw:
Correct.  And I believe only the Queue events won't be admitted.  That is, the `job <message>` events do not rely on the Queue knowing about the job.

BTW, I am considering an alternate approach to solving the problem.  That is, make a separate Redis call to get the new incr ID.  This was suggested in the analysis in #78 .  This entails  the extra overhead of two Redis calls, but has some simplicity, and is much less disruptive vis-a-vis the `checkHealth().newestJob` semantics.

I'm still struggling with test failures in a Docker environment that suggest there are lurking race conditions in some of the test logic (or conceivably the library).

hughsw:
There may be some Queue assumptions that are causing test failures for either approach (this PR or a separate incr call).   The existing behavior is that the Redis state has the saved job in play before the Queue has the job.id in its activeJobs set.  The new behavior is that the Queue has the job.id in its activeJobs set before Redis has the job entered into its state.   Thus, there are different constraints on what can be done...

There's more for me to learn before I can untangle the implications...

hughsw:
See also #194 which implements the Docker test harness I've been using, and documents two intermittent test failures I've seen.

hughsw:
This proposed fix is obsoleted by #197 , so I'm closing it.

PR #195, opened 2020-04-06
feat(test/RedisHost): Use optional env BEE_QUEUE_TEST_REDIS for Redis url during testing

hughsw:
This allows use of a non-localhost Redis server.  It is necessary for the following PR which uses a non-local Redis host in a docker-compose container environment.

Regarding Docker: Not sure how to use PR base from my fork: see https://github.com/hughsw/bee-queue/pull/1  which uses docker-compose to run tests in a Docker  environment.

### DIFF
--- a/test/delay-test.js
+++ b/test/delay-test.js
@@ -38,7 +38,8 @@ function reef(n = 1) {
 }
 
 describe('Delayed jobs', (it) => {
-  const gclient = redis.createClient();
+  const redisUrl = process.env.BEE_QUEUE_TEST_REDIS;
+  const gclient = redis.createClient(redisUrl);
 
   it.before(() => gclient);
 
@@ -55,6 +56,16 @@ describe('Delayed jobs', (it) => {
     });
 
     function makeQueue(...args) {
+      if (redisUrl) {
+        if (args.length === 0) {
+          args.push({});
+        }
+        if (!args[0].redis) {
+          // Note: we don't fuss with isClient(redis) because it's simpler to just
+          // add the host setting in the test code itself when redis is used
+          args[0].redis = redisUrl;
+        }
+      }
       const queue = new Queue(ctx.queueName, ...args);
       queue.on('error', (err) => ctx.queueErrors.push(err));
       ctx.queues.push(queue);
--- a/test/queue-test.js
+++ b/test/queue-test.js
@@ -91,7 +91,10 @@ function spitter() {
 }
 
 describe('Queue', (it) => {
-  const gclient = redis.createClient();
+  const redisUrl = process.env.BEE_QUEUE_TEST_REDIS;
+  // redisUrl must have a schema portion, e.g. redis://redis.example.com
+  const redisHost = redisUrl ? new URL(redisUrl).hostname : undefined;
+  const gclient = redis.createClient(redisUrl);
 
   it.before(() => gclient);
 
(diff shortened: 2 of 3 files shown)


test/delay-test.js
```diff
@@ -36,7 +36,8 @@ function reef(n = 1) {
 }
 
 describe('Delayed jobs', (it) => {
-  const gclient = redis.createClient();
+  const redisHost = process.env.BEE_QUEUE_REDIS_HOST;
```
skeggse, line 39: What do you think about calling this `BEE_QUEUE_TEST_REDIS` and passing it to the `url` parameter of `createClient` for broader compatibility? We tend to prefer using the fully qualified URI so we can include `rediss://` (TLS) or the port.

hughsw: Yes.  My pattern comes from code where we have env vars for each of schema, host, port, endpoint, etc, and only the HOST part got into this code.  Such configurability is not warranted here.  I'll make the proposed change.

hughsw:
The conflict resolution is fairly straightforward.  I've done it locally and also pushed a rebase of the resolved code and additional code to make it work in the Docker setting: https://github.com/hughsw/bee-queue/tree/docker-testing-rb1

test/delay-test.js
```diff
@@ -53,6 +54,16 @@ describe('Delayed jobs', (it) => {
     });
 
     function makeQueue(...args) {
+      if (redisUrl) {
+        if (args.length === 0) {
+          args.push({});
+        }
+        if (!args[0].redis) {
```
skeggse, line 61:
```suggestion
        } else if (!args[0].redis) {
```

hughsw:
It will be a forced push, OK with you?

test/delay-test.js
hughsw: I've made this change in my rebased branch.

hughsw: Actually, I'm not sure this is a correct fix.  Even if no args are provided, we want to use the redisUrl.  And, all uses of `makeQueue()` in this file provide arguments, so the no-arg case is never exercised!  I'm going to leave it as it was because that should provide the least surprise if/when `makeQueue()` is called with no arguments.

skeggse:
>Just to be clear: I will do a forced push to this branch, hughsw:redis-host-env-testing, as rebased on the current bee-queue:master. OK with you?

Yes! Your PR is your PR - force pushes are just fine there.

hughsw:
I've rebased on a recent `master`.  I've applied changes discussed above and added one comment about the schema being required.

test/queue-test.js
```diff
@@ -109,6 +112,16 @@ describe('Queue', (it) => {
     });
 
     function makeQueue(...args) {
+      if (redisUrl) {
+        if (args.length === 0) {
+          args.push({});
+        }
+        if (!args[0].redis) {
```
skeggse, line 119: Will this not `throw` when called with no `args` and a configured `BEE_QUEUE_TEST_REDIS`?

hughsw: I don't think so -- am I missing something?  I'll dig into it tomorrow....

hughsw: I don't get an error.
I added the ` it('reports a succeeded job', async (t) => { ... ` test from `queue-test.js` to `delay-test.js`.  It uses `makeQueue()` with no args, and it runs without error in the configured `BEE_QUEUE_TEST_REDIS` environment I'm using with Docker (and for which I'm about to post a PR).

skeggse: Oh I'm really sorry - I don't know what was happening yesterday but I musta had a bug in my braincompiler. This is right!

PR #196, opened 2020-04-07
feat(example/Docker): Add a Docker example that reproduces Issue 189

hughsw:
This is about the simplest code I can come up with to demonstrate the problem in #189.  

See the README:
https://github.com/hughsw/bee-queue/blob/example-189/examples/docker/issue-189/README.md

### DIFF
--- a/examples/docker/issue-189/.gitignore
+++ b/examples/docker/issue-189/.gitignore
@@ -1,0 +1,1 @@
+/bee-queue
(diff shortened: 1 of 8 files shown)


hughsw:
I am working on a throughput example that will demonstrate the issue if it hasn't been resolved.  So, this PR is obsolete and I'm closing it.

PR #197, opened 2020-04-07
Fix dropped jobs issue

hughsw:
Changes to fix problem reported in #189 and #78 

This fix adopts an approach mentioned in #78 of making a separate Redis call to updated the id counter.  Some simplifications emerge.

The example code in #196 shows no lost jobs when run against this branch.

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -725,6 +725,7 @@ class Queue extends Emitter {
       .then((results) => {
         const numRaised = results[0], nextOpportunity = results[1];
         if (numRaised) {
+          // not documented
           this.emit('raised jobs', numRaised);
         }
         this._delayedTimer.schedule(parseInt(nextOpportunity, 10));
--- a/test/queue-test.js
+++ b/test/queue-test.js
@@ -342,7 +342,7 @@ describe('Queue', (it) => {
         const jobs = spitter();
         queue.process((job) => jobs.pushSuspend(job));
 
-        await queue.createJob({}).save();
+        const job = await queue.createJob({}).save();
         const [, finishJob] = await jobs.shift();
 
         await t.throws(queue.close(10));
@@ -352,7 +352,7 @@ describe('Queue', (it) => {
 
         const errors = t.context.queueErrors, count = errors.length;
         t.context.queueErrors = errors.filter((err) => {
-          return err.message !== 'unable to update the status of succeeded job 1';
+          return err.message !== `unable to update the status of succeeded job ${job.id}`;
         });
         t.is(t.context.queueErrors.length, count - 1);
         t.context.handleErrors(t);
@@ -878,36 +878,35 @@ describe('Queue', (it) => {
   });
 
   it.describe('removeJob', (it) => {
-    it('should not cause an error if immediately removed', async (t) => {
+
+    it('should not cause an error if removed before started', async (t) => {
+      // creates three jobs, waits until all are saved, then processes, first job removes the second one
+
       const queue = t.context.makeQueue();
 
+      const wait = helpers.deferred(), finishWait = wait.defer();
+
+      await queue.createJob({foo: 'bar'}).setId('goodjob1').save();
+      const deadjob = await queue.createJob({foo: 'bar'}).setId('deadjob').save();
+      await queue.createJob({foo: 'bar'}).setId('goodjob2').save();
+
       queue.process(async (job) => {
+        await wait;
+        if (job.id === 'goodjob1') {
+          t.is(deadjob.id, 'deadjob');
+          await queue.removeJob(deadjob.id);
+        }
         if (job.id === 'deadjob') {
-          t.fail('should not be able to process the job');
+          t.fail('should not get to process deadjob');
         }
       });
 
-      const waitJob = queue._waitForJob, wait = helpers.deferred();
-      let waitDone = wait.defer();
-      queue._waitForJob = function (...args) {
-        if (waitDone) {
-          waitDone();
-          waitDone = null;
-        }
-        return waitJob.apply(this, args);
-      };
+      finishWait();
 
-      await wait;
-
-      const job = queue.createJob({foo: 'bar'}).setId('deadjob');
-      await Promise.all([
-        job.save(),
-        queue.removeJob(job.id),
-        queue.createJob({foo: 'bar'}).setId('goodjob').save(),
-      ]);
-
-      const goodJob = await helpers.waitOn(queue, 'succeeded');
-      t.is(goodJob.id, 'goodjob');
+      const succeeded1 = await helpers.waitOn(queue, 'succeeded');
+      t.is(succeeded1.id, 'goodjob1');
+      const succeeded2 = await helpers.waitOn(queue, 'succeeded');
+      t.is(succeeded2.id, 'goodjob2');
 
       t.context.handleErrors(t);
     });
@@ -1163,7 +1162,7 @@ describe('Queue', (it) => {
       const [[failedJob, err]] = fail.args;
 
       t.truthy(failedJob);
-      t.is(job.id, '1');
+      t.is(job.id, failedJob.id);
       t.is(failedJob.data.foo, 'bar');
       t.is(err.message, `Job ${job.id} timed out (10 ms)`);
     });
--- a/lib/job.js
+++ b/lib/job.js
@@ -50,34 +50,48 @@ class Job extends Emitter {
     const toKey = this.queue.toKey.bind(this.queue);
 
     let promise;
+    if (this.id) {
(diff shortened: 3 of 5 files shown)


hughsw:
I've closed #193 which had an alternate approach, and which has considerable discussion.

hughsw:
I will have to look into the decrease in code coverage.  Very likely due to new branches in `Job.save()`

hughsw:
Unfortunately the old version of "should not cause an error if immediately removed" fails when run with the new id-setting logic.

hughsw:
By a gross, but deterministic hack to `Queue._waitForJob()` I can make this test pass.  The success of the hack supports my contention that the test was really only working due to fortuitous race conditions rather than a deterministic flow of control.

hughsw:
Evidence:
Prior without these changes, `master` shows over a 2% job loss rate as per the Docker exampe #196
```
client_1  | {"numJobSaveSuccess":3588,"numJobSaveError":0,"numQueueSucceeded":3586,"numQueueFailed":0,"numJobSucceeded":3461,"throughput":"1195","numSucceededLost":125,"succededLossPercent":"3.5"}
client_1  | {"numJobSaveSuccess":9154,"numJobSaveError":0,"numQueueSucceeded":9155,"numQueueFailed":0,"numJobSucceeded":8901,"throughput":"1526","numSucceededLost":254,"succededLossPercent":"2.8"}
client_1  | {"numJobSaveSuccess":16538,"numJobSaveError":0,"numQueueSucceeded":16544,"numQueueFailed":0,"numJobSucceeded":16180,"throughput":"1838","numSucceededLost":364,"succededLossPercent":"2.2"}
client_1  | {"numJobSaveSuccess":24535,"numJobSaveError":0,"numQueueSucceeded":24532,"numQueueFailed":0,"numJobSucceeded":24017,"throughput":"2044","numSucceededLost":515,"succededLossPercent":"2.1"}
```

With the changes of this PR, after running overnight, I have 0 lost jobs from over 44 million jobs run:
```
client_1  | {"numJobSaveSuccess":44795256,"numJobSaveError":0,"numQueueSucceeded":44795249,"numQueueFailed":0,"numJobSucceeded":44795249,"throughput":"1226","numSucceededLost":0,"succededLossPercent":"0.0"}
client_1  | {"numJobSaveSuccess":44798380,"numJobSaveError":0,"numQueueSucceeded":44798370,"numQueueFailed":0,"numJobSucceeded":44798370,"throughput":"1226","numSucceededLost":0,"succededLossPercent":"0.0"}
client_1  | {"numJobSaveSuccess":44801830,"numJobSaveError":0,"numQueueSucceeded":44801821,"numQueueFailed":0,"numJobSucceeded":44801821,"throughput":"1226","numSucceededLost":0,"succededLossPercent":"0.0"}
```

lib/queue.js
```diff
@@ -420,11 +420,20 @@ class Queue extends Emitter {
...
 
-  _waitForJob() {
-    const idPromise = helpers.deferred();
+  _waitForJob(testingCb) {
+    let idPromise = helpers.deferred();
     this.bclient.brpoplpush(this.toKey('waiting'), this.toKey('active'), 0,
       idPromise.defer());
 
+    // The only reason for this callback is to support deterministic testing of
+    // removeJob between the time brpoplpush resolves and Job.fromId() runs
```
skeggse, line 429: Instead of modifying the source to support the test, what do you think about modifying the test to either stub the `Job.fromId` static method or modify the `brpoplpush` function to provide this hook?

lib/queue.js
```diff
@@ -725,6 +734,7 @@ class Queue extends Emitter {
       .then((results) => {
         const numRaised = results[0], nextOpportunity = results[1];
         if (numRaised) {
+          // not documented
```
skeggse, line 737: heh yeah

This seems unrelated to the overarching PR, and maybe we could document it instead of leaving it more explicitly undocumented.

test/queue-test.js
```diff
@@ -460,12 +460,12 @@ describe('Queue', (it) => {
...
       const waitJob = queue._waitForJob, wait = helpers.deferred();
       let waitDone = wait.defer();
-      queue._waitForJob = function (...args) {
+      queue._waitForJob = function () {
         if (waitDone) {
           waitDone();
           waitDone = null;
         }
-        return waitJob.apply(this, args);
+        return waitJob.apply(this);
```
skeggse, line 468: `apply` like this just always feels weird to me :\
```suggestion
        return waitJob.call(this);
```

hughsw:
Another solution I had working was for `Job.save()` to generate a uuid.  This avoids the extra Redis call.  It intrudes slightly on the semantics of `newestJob` from `Queue.checkHealth()`.

Regarding #198 -- I imagine the present implementation of `Queue.saveAll()` will make it very easy to demonstrate the dropped jobs problem because so much time passes before the Queue gets to update its jobs table.

Do you think I should I pursue the uuid approach with an eye towards how `Queue.saveAll()` should be implemented?

lib/queue.js
hughsw: Yes, I had thought about wrapping `Job.fromId()` -- I'll revisit that approach.

test/queue-test.js
hughsw: You're right.

skeggse:
>Regarding #198 -- I imagine the present implementation of Queue.saveAll() will make it very easy to demonstrate the dropped jobs problem because so much time passes before the Queue gets to update its jobs table.

That's possible, though my (albeit fairly ancient) understanding of how pipelined commands work is that Redis still iterates through clients and accepts commands one at a time from each client where there are buffered commands. What that means (I think) is that the `saveAll` call will put a bunch of save commands in the kernel buffer on the `redis-server`'s side, and redis will read one of them, then read a command from another client (e.g. the worker client). Worth testing, though! I could imagine adding a semi-atomic mode to `saveAll` that uses `multi` instead of `batch` to more or less do what you're suggesting.

---

I like the approach in #193 due to its simplicity and its superior scaling properties/avoidance of additional latency in the save-job flow.

I think the key requirements that I'm looking for in a fix for #78 / #198 are (specifically when `storeJobs` and `getEvents` are `false` - I can imagine solutions that change the behavior in other cases):

- Does not alter the default job ID format, at least prior to bee-queue 2.0
- Does not introduce additional network latency by virtue of making additional Redis calls that require sequential feedback (i.e. make a redis call to get the ID, then make a redis call to save the job)

So based on those requirements, here are my thoughts (in order of preference, I think):

- Add an extra redis call in some cases (a la this PR), either by default when necessary to support the desired features (event delivery), or by explicit opt-in
- Change the format of the ID in some cases (a la #193), either by default when necessary to support the desired features (event delivery), or by explicit opt-in (which could become default in bee-queue 2.0)
  - Possibly also add a new key to the data model to track completed and accepted jobs, for metrication
  - Possibly learn from MongoDB's ObjectID scheme that attempts to reduce the (already incredibly small) possibility of ID collision
- Augment the data model to store jobs that have completed but for whom the sender has not yet acknowledged the completion - not sure what would inspire the sender to go and check this (maybe a periodic check), but it would allow eventual event delivery at the expense of additional complexity and higher latency in the exceptional case
- Augment the data model to use redis streams (one per client/sender, for example; might use an alternate pull-messaging approach per [here](https://redislabs.com/ebook/part-2-core-concepts/chapter-6-application-components-in-redis/6-5-pull-messaging/)) for event delivery instead of pubsub (possibly as a configurable feature); this might be a breaking change, and would introduce concerns around client persistence and the delivery of the updates to non-senders
- Augment the pubsub approach to publish using a key that includes the job ID. In the completion script, if the `publish` doesn't see any receivers for the event, it uses the augmented data model to store the job's final status for eventual delivery
- Add a pluggable job ID allocation mechanism (along the lines of a plugin system)
- Reserve blocks of IDs for a given worker, so that it can synchronously dole out IDs while mostly preserving the information in the `id` key for input metrication
  - This adds a lot of complexity, so not a preferred solution
- Add an artificial delay to the publish in `_finishJob` if the job finishes quickly. Not a complete solution, as it only mitigates the race condition - it does not solve it - and it hurts bee-queue's core features of performance and latency

I suspect the solution will be one of the first two, with the thinking that the event delivery system will become a plugin at some point. Ultimately, the solution will need to be predicated on your needs, as we don't use bee-queue's event delivery due to its somewhat suboptimal scaling properties (and lack of need for its functionality to begin with).

hughsw:
There's a lot to think about that you've outlined.  I appreciate your putting it all down.  We've touched on version 2 in other PR and Issues threads.  How much of the API is on the table for version 2?  Where is that discussion happening?  I have some radical simplification ideas, prompted by your mentions of a more plug-in centric architecture.

PR #198, opened 2020-04-08
feat: implement Queue#saveAll feature

skeggse:
This feature pipelines the `EVALSHA` commands for many `Job#save` calls together to avoid the compounding latency penalty when using the `redis` library.

### DIFF
--- a/test/queue-test.js
+++ b/test/queue-test.js
@@ -963,6 +963,118 @@ describe('Queue', (it) => {
     });
   });
 
+  it.describe('saveAll', (it) => {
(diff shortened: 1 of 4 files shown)


test/queue-test.js
```diff
@@ -877,6 +877,88 @@ describe('Queue', (it) => {
     });
   });
 
+  it.describe('saveAll', (it) => {
```
shils, line 880: Mind adding a test that shows when `saveAll` rejects?

skeggse: Done!

PR #215, opened 2020-04-10
fix: misc edge case fixes

skeggse:
This is an incremental improvement, mostly.

Fixes:

- produce uncaught exceptions instead of unhandled rejections
- invoke `Queue#close` callback in edge case
- improve error reporting from `Queue#close`
- emit `error` events from `Queue` when un-stall command fails
- fix checkStalledJobs error handling
- make checkStalledJobs to recurse even after a failure
- prevent errant uncaught exceptions due to errors early in client connection

Codebase quality:

- add prettier
- add eslint rules
- improve testing, fix some race conditions
- upgrade outdated modules
- use callAsync over long-form deferred
- some refactoring


package.json
```diff
@@ -4,6 +4,7 @@
   "description": "A simple, fast, robust job/task queue, backed by Redis.",
   "main": "index.js",
   "dependencies": {
+    "p-finally": "^1.0.0",
     "promise-callbacks": "^3.0.0",
```
shils, line 8: This change needs at least v3.4.0 to work as intended- https://github.com/bee-queue/bee-queue/pull/215/commits/0cfba99956b700c420852e9f342528ab8b8b1da4

lib/queue.js
```diff
@@ -199,28 +217,36 @@ class Queue extends Emitter {
...
+            // Wait for all the jobs to complete. Ignore job errors during
+            // shutdown.
+            return Promise.all(
+              Array.from(this.activeJobs, (promise) => promise.catch(() => {}))
+            );
+          }),
+          timeout
+        )
+        .then(() => {}),
+      () => cleanup()
```
shils, line 248: Potential unhandled rejection?

skeggse: `finally` actually waits for any `Promise` returned from the provided function!

lib/queue.js
```diff
@@ -606,92 +666,103 @@ class Queue extends Emitter {
...
+      return stalled.length;
+    });
+  }
+
+  _safeCheckStalledJobs(interval, cb) {
+    const promise = this._checkStalledJobs(interval, cb);
+    // If a callback is not defined, then we must emit errors to avoid unhandled
+    // rejections. If there is a callback, then _checkStalledJobs will attach it
+    // as an error handler to `promise`.
+    if (!cb) promise.catch(this._emitErrorAfterTick);
```
shils, line 740: Any drawback to attaching this in `_checkStalledJobs` itself?

skeggse: Yeah - that'd result in the `Queue` emitting an `error` event even in the case that we're calling `checkStalledJobs` directly and handling the returned `Promise`. This only exists for subsequent calls on the originally specified interval.

package.json
skeggse: 9c456e1a5dc619f38d67b3ab3d92cce3b0d162bc

lib/queue.js
skeggse: Resolved offline - solution was bbb56903828e4d22875dfa90d4f720ee27c47c29, and we don't anticipate the swallowing of subsequent errors to be problematic.

lib/queue.js
```diff
@@ -222,7 +222,7 @@ class Queue extends Emitter {
       this._isClosed = true;
 
       const clients = [];
-      if (this.settings.quitCommandClient) {
+      if (this.settings.quitCommandClient && this.client) {
```
bradvogel, line 225: worth unit testing this?

lib/helpers.js
```diff
@@ -6,15 +6,32 @@ function has(object, name) {
   return hasOwn.call(object, name);
 }
 
+/**
+ * A variant of the Promise#finally implementation. Instead of rejecting with
+ * the error that occurs in the finally clause, it rejects with the error from
+ * the original Promise first, and falls back to using the error from the
+ * finally clause if no such error occurred.
+ */
+function finallyRejectsWithInitial(promise, fn) {
```
bradvogel, line 15: worth unit testing?

lib/queue.js
skeggse: Good call! Found another bug :) 3183887d50d6c37533223a6eeecb17cac003e859

lib/helpers.js
skeggse: bb2b9153eab3d4deb025d0d0af173e9ac5fb7d5d

bradvogel:
awesome, thanks! I didn't look thoroughly for other missing unit test coverage - I trust that you and/or the tooling will uncover that.

PR #229, opened 2020-04-15
docs(examples): Add Hello World example for Bee-Queue in Docker containers

hughsw:
A minimalist Hello World for Bee-Queue using Docker containers in a networked setting with client and worker running in different Node JS processes.  The main goal of this very terse example is for the user to vet their Docker framework.  Once this is merged I will put up other PRs with more detailed and informative examples.

The Readme:  https://github.com/hughsw/bee-queue/blob/hello-world-docker-example/examples/docker/00-hello-world/Readme.md

### DIFF
--- a/examples/docker/00-hello-world/client.js
+++ b/examples/docker/00-hello-world/client.js
@@ -1,0 +1,28 @@
+const Queue = require('bee-queue');
+
+// Core client logic, run one job and then close the queue
+const clientProcess = (queue) => {
+  const job = queue.createJob({name: 'World'});
+  job.on('succeeded', (result) => {
+    console.log();
+    console.log(result.greeting);
+    console.log();
+
+    queue.close();
+  });
+
+  job.save();
+};
+
+const go = () => {
+  return new Queue('hello-world', {redis: {host: 'redis'}, isWorker: false})
+    .ready()
+    .then(clientProcess)
+    .catch((err) =>
+      process.nextTick(() => {
+        throw err;
+      })
+    );
+};
+
+go();
--- a/examples/docker/00-hello-world/go.sh
+++ b/examples/docker/00-hello-world/go.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+# Build and run bee-queue Docker containers to do Hello World
+
+# Strict and loud fail
+set -euo pipefail
+trap 'rc=$?;set +ex;if [[ $rc -ne 0 ]];then trap - ERR EXIT;echo 1>&2;echo "*** fail *** : code $rc : $DIR/$SCRIPT $ARGS" 1>&2;echo 1>&2;exit $rc;fi' ERR EXIT
+ARGS="$*"
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+SCRIPT="$(basename "${BASH_SOURCE[0]}")"
+
+echo
+docker --version
+docker-compose --version
+echo
+
+set -x
+docker-compose down > /dev/null 2>&1 || true
+docker-compose build --quiet
+docker-compose up --abort-on-container-exit --timeout 3 --renew-anon-volumes
+set +x
--- a/examples/docker/00-hello-world/worker.js
+++ b/examples/docker/00-hello-world/worker.js
@@ -1,0 +1,19 @@
+const Queue = require('bee-queue');
+
+// Core worker logic, return an object with a job.data-based greeting
+const workerProcess = async (job) => {
+  return {greeting: 'Hello ' + job.data.name + '!'};
+};
+
+const go = () => {
+  return new Queue('hello-world', {redis: {host: 'redis'}, isWorker: true})
+    .ready()
+    .then((queue) => queue.process(workerProcess))
+    .catch((err) =>
+      process.nextTick(() => {
+        throw err;
+      })
+    );
+};
+
+go();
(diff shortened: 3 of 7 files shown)


examples/docker/00-hello-world/worker.js
```diff
@@ -0,0 +1,15 @@
...
+// Core worker logic
+const workerProcess = (job, done) => {
+  done(null, {greeting: 'Hello ' + job.data.name + '!'});
+};
+
+const go = () => {
+  return new Queue('hello-world', {redis: {host: 'redis'}, isWorker: true})
+    .ready()
+    .then((queue) => queue.process(workerProcess))
+    .catch(console.error);
```
skeggse, line 12: I'd recommend failing hard instead of just logging the error.
```suggestion
    .catch((err) =>
      process.nextTick(() => {
        throw err;
      });
    );
```

examples/docker/00-hello-world/worker.js
```diff
@@ -0,0 +1,15 @@
+const Queue = require('bee-queue');
+
+// Core worker logic
+const workerProcess = (job, done) => {
+  done(null, {greeting: 'Hello ' + job.data.name + '!'});
+};
```
skeggse, line 6: It might be preferable to make this an `async` function, as this is an example
```suggestion
async function workerProcess(job) {
  return {greeting: 'Hello ' + job.data.name + '!'};
}
```

examples/docker/00-hello-world/go.sh
```diff
@@ -0,0 +1,13 @@
...
+
+echo
+docker --version
+docker-compose --version
+echo
+
+set -x
+docker-compose down > /dev/null 2>&1 || true
+docker-compose build  --quiet
+docker-compose up  --abort-on-container-exit --timeout 3 --renew-anon-volumes
```
skeggse, line 13:
```suggestion
docker-compose build --quiet
docker-compose up --abort-on-container-exit --timeout 3 --renew-anon-volumes
```

examples/docker/00-hello-world/client.js
```diff
@@ -0,0 +1,24 @@
...
+  });
+
+  job.save();
+};
+
+const go = () => {
+  return new Queue('hello-world', {redis: {host: 'redis'}, isWorker: false})
+    .ready()
+    .then(clientProcess)
+    .catch(console.error);
```
skeggse, line 21: ditto on the fail-hard here

examples/docker/00-hello-world/client.js
```diff
@@ -0,0 +1,24 @@
...
+
+// Core client logic, run one job and then exit
+const clientProcess = (queue) => {
+  const job = queue.createJob({name: 'World'});
+  job.on('succeeded', (result) => {
+    console.log();
+    console.log(result.greeting);
+    console.log();
+
+    process.exit(0);
```
skeggse, line 11: I usually prefer to let the event loop empty vs forcibly exiting. What do you think?

maybe with some error handling:
```suggestion
    queue.close();
```

examples/docker/00-hello-world/go.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
skeggse, line 1:
```suggestion
#!/bin/bash -e
```

examples/docker/00-hello-world/worker.js
hughsw: Yes, much preferred.  I originally had an async using fat-arrow syntax, and the linter choked on it.  These days I refuse to use the `function` keyword ;-) / 2.  Perhaps we could have more relaxed (modern) linter configuration for the examples directories?

examples/docker/00-hello-world/client.js
hughsw: Duh, thank you!

examples/docker/00-hello-world/go.sh
hughsw: I'll put in my standard bash strict boilerplate

examples/docker/00-hello-world/worker.js
hughsw: Heh, the linter doesn't like this syntax either!

examples/docker/00-hello-world/go.sh
hughsw: Done

examples/docker/00-hello-world/worker.js
hughsw: I've reverted the commit until I figure out how to modernize the linter config for examples.

examples/docker/00-hello-world/client.js
hughsw: Done

examples/docker/00-hello-world/worker.js
skeggse: >These days I refuse to use the `function` keyword ;-)

Any particular reason? I much prefer it for named functions, because the syntax is just so easy to scan. For callbacks I definitely prefer the fat-arrow notation.

>Perhaps we could have more relaxed (modern) linter configuration for the examples directories?

Yep!

hughsw: The large Typescript project I'm working on doesn't use the `function` keyword, so it seems somewhat anachronistic when I see it! ;-) / 2   Perhaps it's slightly easier to scan... perhaps.  But also more to scan, and it brings up the dynamic scoping of `this`...   Different styles...

compwright:
Thank you for this submission. A starter project template with Bee-Queue and Docker might be useful to someone who wasn't sure how to set up their project, however that would be better suited for a dedicated repository.

PR #231, opened 2020-04-16
fix(queue): remove error event listener on close

### LINKED ISSUE #102: MaxListenersExceededWarning for 'error' and 'ready' listeners
When creating 11 producer queues with the following configuration I'm seeing the MaxListenersExceededWarning  in node 8.9.1.

> (node:25043) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 error listeners added. Use emitter.setMaxListeners() to increase limit
> (node:25043) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 ready listeners added. Use emitter.setMaxListeners() to increase limit
> (node:25043) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 error listeners added. Use emitter.setMaxListeners() to increase limit

These queues are only meant to be producer queues, to push new jobs into the queue, not worker queues, so the overhead should be minimized.

I _think_ this happens because of two things:

1. the `Queue` class is a subclass of `Emitter` https://github.com/bee-queue/bee-queue/blob/master/lib/queue.js#L13
1. somewhere in the `Queue` instantiation process bee-queue is attaching event handlers to the 'ready' and 'error' events.

Things I can't figure out:

1. Where is number 2 happening?
1. Is `Queue` correctly removing event handlers when necessary
1. Is this a warning that is expected to show up?

It seems reasonable that on a webserver a user of bee-queue would want to instantiate > 10 producer queues. In this reasonable usage of the library I would not expect to get a node warning.

```javascript
const Redis = require('redis')
const Queue = require('bee-queue')
const redis = Redis.createClient('redis://localhost:6379')

const sharedConfig = {
  redis,
  getEvents: false,
  isWorker: false
}

const queueNames = ['queue-1', 'queue-2', 'queue-3',  'queue-4', 'queue-5', 'queue-6', 'queue-7', 'queque-8', 'queue-9', 'queue-10', 'queue-11']
queueNames.forEach((queueName) => {
   const queue = new Queue(queueName, sharedConfig)
   console.log('queue created', queue.name)
})
```

image 2018-03-14 at 11 00 33 am

skeggse:
The error event listener is useful for catching errors from redis clients that aren't managed by the consumer of the library, but might cause problems for re-used connections.

Fixes #102 (per the extended discussion) cc @dylanjha.

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -75,6 +75,8 @@ class Queue extends Emitter {
       return redis
         .createClient(this.settings.redis, createNew)
         .then((client) => {
+          // This event gets cleaned up and removed in Queue#close for the
+          // primary client if quitCommandClient is disabled.
           client.on('error', this._emitError);
           return (this[clientName] = client);
         });
@@ -222,8 +224,12 @@ class Queue extends Emitter {
       this._isClosed = true;
 
       const clients = [];
-      if (this.settings.quitCommandClient && this.client) {
-        clients.push(this.client);
+      if (this.client) {
+        if (this.settings.quitCommandClient) {
+          clients.push(this.client);
+        } else {
+          this.client.removeListener('error', this._emitError);
+        }
       }
       if (this.eclient) {
         clients.push(this.eclient);


dylanjha:
nice. good find

PR #232, opened 2020-04-16
fix: prevent event delivery race condition

### LINKED ISSUE #22: Add tests to verify that job status fields update
Ref #19

### LINKED ISSUE #189: Reproducible: bee-queue loses track of jobs

skeggse:
This solution introduces a new client-specific identifier that can be used independent of the normal identifier. This approach permits the continued use of the global incrementing queue `id` so that we don't break existing requirements around the use of that `id` for reporting and metrics.

Fixes #189, adds tests to fix #22.

### DIFF
--- a/benchmark/bq/harness.js
+++ b/benchmark/bq/harness.js
@@ -16,14 +16,13 @@ const queue = new Queue(
 
 // A promise-based barrier.
 function reef(n = 1) {
-  const done = helpers.deferred(),
-    end = done.defer();
+  const done = helpers.defer();
   return {
-    done,
+    done: done.promise,
     next() {
       --n;
       if (n < 0) return false;
-      if (n === 0) end();
+      if (n === 0) done.resolve();
       return true;
     },
   };
--- a/benchmark/bull/harness.js
+++ b/benchmark/bull/harness.js
@@ -4,14 +4,13 @@ const queue = new Queue('test');
 
 // A promise-based barrier.
 function reef(n = 1) {
-  const done = helpers.deferred(),
-    end = done.defer();
+  const done = helpers.defer();
   return {
-    done,
+    done: done.promise,
     next() {
       --n;
       if (n < 0) return false;
-      if (n === 0) end();
+      if (n === 0) done.resolve();
       return true;
     },
   };
--- a/benchmark/kue/harness.js
+++ b/benchmark/kue/harness.js
@@ -4,14 +4,13 @@ const queue = kue.createQueue();
 
 // A promise-based barrier.
 function reef(n = 1) {
-  const done = helpers.deferred(),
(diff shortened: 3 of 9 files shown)


skeggse:
This is my proposal for fixing the behavior seen in #189 and #78, as it solves some of my concerns per https://github.com/bee-queue/bee-queue/pull/197#issuecomment-614241270. It builds upon both #193 and the discussion in #197

hughsw:
This seems like multiple PRs, with improved error handling, new tests?, fixing a bug, and client UUID support mingled together.  Being new to the code base it will take me some time to separate the concerns and their implementations.

The improved error handling seems like a good thing.  At first blush the client UUID support seems heavy, lots of new code and conditionals -- are there requests for this?

skeggse:
>This seems like multiple PRs

That's...a running problem for me. I'll look at splitting this up into more digestible PRs. Sometimes I start working on stuff and the scope just keep creeping. Sorry about that!

>At first blush the client UUID support seems heavy, lots of new code and conditionals -- are there requests for this?

It's specifically to maintain feature parity for our internal tooling we've built around bee-queue, yeah.

hughsw:
Yeah, I'm familiar with the feature creep on a dev branch.

Can you say more about "feature parity" details that need to be maintained.  I can think of three:
1. `Queue#checkHealth().newestJob` counts the number of default-id jobs
2. `Job.save()` makes a single Redis call (or, preserve benchmark performance)
3. `job.id` have a lexical ordering such that job IDs sort according to the order in which they were saved

And of course, for the bug (#78) to be fixed, all successfully saved jobs will emit exactly one of `'succeeded'` or `'failed'` event

Are there other features or semantics that need to be preserved or have been requested?

Of the three PRs approaching these matters:
- #193 (obsolete) Juggled all three features over various commits, but never had all three
- #232 Preserves all three, and adds several other refactorings and new features, I believe (it's hard to tease apart all the threads of changes)
- #237 Preserves (1) and (2), and PR commentary points out a simple way to preserve (3) to within 1 msec

Does this line of questioning and feature factoring make sense?

PR #233, opened 2020-04-17
fix(prettier): Fix prettier glob  for file extensions in package.json

hughsw:
My work with building and testing Bee-Queue with Docker uncovered this problem with the prettier glob.


hughsw:
Heh, you beat me to applying the --write fixes!

PR #234, opened 2020-04-17
[Obsolete draft] dev(testing/docker): Add support for local, Docker-based testing

hughsw:
**[ This PR is made obsolete by #239 ]**

See Testing section at bottom of README: https://github.com/hughsw/bee-queue/blob/docker-testing-sq1/README.md#testing

### DIFF
--- a/Dockerfile-tests
+++ b/Dockerfile-tests
@@ -1,0 +1,46 @@
+# We build an image with bee-queue's node_modules and all the source
+# files.  We jump through some Docker FROM hoops to avoid having the
+# bloat of the g++ compiler and other tools in the final image.
+
+FROM node:lts-alpine as beequeue-base
+
+# Info, and add bash, which we need in the final image
+RUN true \
+  && echo -n "node " && node --version \
+  && echo -n "npm " && npm --version \
+  && echo -n "apk " && apk --version \
+  && apk update \
+  && apk add --no-cache bash \
+  && rm -r /usr/share/man \
+  && true
+
+FROM beequeue-base as beequeue-node_modules
+
+# Ensure that tools for npm are available (for node-gyp in particular)
+# to build node_modules
+RUN true \
+  && apk add --no-cache python make g++ \
+  && true
+
+WORKDIR /bee-queue
+
+# Start with just these two files so as to optimize that npm install
+# only runs when package.json has been updated
+COPY package.json package-lock.json ./
+RUN npm --version && npm ci
+
+# Build the target image
+FROM beequeue-base
+
+WORKDIR /bee-queue
+
+COPY --from=beequeue-node_modules /bee-queue/node_modules ./node_modules/
+
+# We use a script for Docker-based testing (instead of npm test) so
+# that the dev cycle doesn't have to wait for the npm install
+# triggered by editing e.g. the scripts stanza in package.json
+ENV TEST_SCRIPT $TEST_SCRIPT
+CMD ./${TEST_SCRIPT}
+
+# Now copy all the source files
+COPY . ./
--- a/tests.sh
+++ b/tests.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+# Default script used by the Docker testing harness in the container run by ./docker-tests.sh
+
+# Strict, and loud failure
+set -euo pipefail
+trap 'rc=$?;set +ex;if [[ $rc -ne 0 ]];then trap - ERR EXIT;echo 1>&2;echo "*** fail *** : code $rc : $DIR/$SCRIPT $ARGS" 1>&2;echo 1>&2;exit $rc;fi' ERR EXIT
+ARGS="$*"
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+SCRIPT="$(basename "${BASH_SOURCE[0]}")"
+
+set -x
+
+# Dev-centric test steps based on scripts in package.json
+PATH=$(npm bin):$PATH
+
+# If ava testing hangs (which, sadly, happens all too often), add --serial to get a more
+# deterministic record that will log up to the test prior to the one that hangs... But, as
+# is often the case with race conditions, the instrumentation may eliminate the problem
+# you're trying to track down...
+ava --no-color --timeout 30000 --verbose
+eslint .
+prettier --check '**/*.(html|json|md|sublime-project|ts|yml)'
+nyc ava
+
+# To do what Travis does, but, to date, the ava tests will intermittently hang
+#npm test
--- a/tests.sh
+++ b/tests.sh
@@ -12,7 +12,7 @@ SCRIPT="$(basename "${BASH_SOURCE[0]}")"
 set -x
 
 # Dev-centric test steps based on scripts in package.json
-PATH=$(npm bin):$PATH
+PATH="$(npm bin):$PATH"
 
 # If ava testing hangs (which, sadly, happens all too often), add --serial to get a more
 # deterministic record that will log up to the test prior to the one that hangs... But, as
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+/.git
+/.nyc_output
+/node_modules
--- a/README.md
+++ b/README.md
@@ -882,7 +882,11 @@ Some significant non-features include:
 
 Some of these could be worthwhile additions; please comment if you're interested in using or helping implement them!
 
-You'll need a local redis server to run the tests. Note that running the tests may delete some keys in the form of `bq:test-*-*:*`.
+## Testing
+
+Running tests via `npm test` needs an non-password Redis server at port 127.0.0.1 to run the tests. Note that the testing will delete some Redis keys with the pattern `bq:test-*-*:*`.
+
+Alternatively, if you have Docker and docker-compose installed, you can build an image and run tests via `./docker-tests.sh`. This approach uses an ephemeral Redis server each time it is run. Be patient the first time you run it while Docker is downloading and populating its caches, and also each time package.json has been modified. The `./tests.sh` script is what's run in the container. Or, you can provide your own script as an argument to `./docker-tests.sh`.
 
 [npm-image]: https://img.shields.io/npm/v/bee-queue.svg?style=flat
 [npm-url]: https://www.npmjs.com/package/bee-queue
--- a/bee-queue.sublime-project
+++ b/bee-queue.sublime-project
@@ -8,9 +8,7 @@
         "coverage",
         ".nyc_output"
       ],
-      "file_exclude_patterns": [
-        "package-lock.json"
-      ]
+      "file_exclude_patterns": ["package-lock.json"]
     }
   ],
   "settings": {
--- a/docker-compose-tests.yml
+++ b/docker-compose-tests.yml
@@ -1,0 +1,17 @@
+version: '3.7'
+
+services:
+  bee-queue:
+    build:
+      context: .
+      dockerfile: Dockerfile-tests
+    environment:
+      # Url for the Redis server, must have schema
+      - BEE_QUEUE_TEST_REDIS=redis://redis
+      - TEST_SCRIPT
+
+  redis:
+    image: redis:alpine
+    # sigh, escaping to get the shell to echo what it's running
+    command: /bin/sh -c set\ -x\ &&\ redis-server\ --tcp-backlog\ 128\ --logfile\ /dev/null
+    #command: redis-server --tcp-backlog 128 --logfile /dev/null
--- a/docker-tests.sh
+++ b/docker-tests.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+
+# Build a Docker image and run tests in a container on a virtual network with an ephemeral
+# Redis server container. Run ./tests.sh by default.  A user script can be provided in $1
+
+# Strict, and loud failure
+set -euo pipefail
+trap 'rc=$?;set +ex;if [[ $rc -ne 0 ]];then trap - ERR EXIT;echo 1>&2;echo "*** fail *** : code $rc : $DIR/$SCRIPT $ARGS" 1>&2;echo 1>&2;exit $rc;fi' ERR EXIT
+ARGS="$*"
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+SCRIPT="$(basename "${BASH_SOURCE[0]}")"
+
+cd $DIR
+
+# Use tests.sh by default, or a user-provided script
+export TEST_SCRIPT="${1:-tests.sh}"
+if [[ !(-f "$TEST_SCRIPT" && -x "$TEST_SCRIPT") ]] ; then
+    echo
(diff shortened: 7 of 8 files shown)


tests.sh
```diff
@@ -0,0 +1,27 @@
...
+set -euo pipefail
+trap 'rc=$?;set +ex;if [[ $rc -ne 0 ]];then trap - ERR EXIT;echo 1>&2;echo "*** fail *** : code $rc : $DIR/$SCRIPT $ARGS" 1>&2;echo 1>&2;exit $rc;fi' ERR EXIT
+ARGS="$*"
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+SCRIPT="$(basename "${BASH_SOURCE[0]}")"
+
+set -x
+
+# Dev-centric test steps based on scripts in package.json
+PATH=$(npm bin):$PATH
```
skeggse, line 15:
```suggestion
PATH="$(npm bin):$PATH"
```

Dockerfile-tests
```diff
@@ -0,0 +1,46 @@
...
+# Build the target image
+FROM beequeue-base
+
+WORKDIR /bee-queue
+
+COPY --from=beequeue-node_modules /bee-queue/node_modules ./node_modules/
+
+# We use a script for Docker-based testing (instead of npm test) so
+# that the dev cycle doesn't have to wait for the npm install
+# triggered by editing e.g. the scripts stanza in package.json
```
skeggse, line 41: I'm not sure I follow this - if I'm not mistaken, `npm test` just executes the commands defined in the `test` script - how does `npm install` factor into it? I'd like to avoid duplicating the commands defined in the `package.json` if at all possible.

tests.sh
```diff
@@ -0,0 +1,27 @@
...
+
+# If ava testing hangs (which, sadly, happens all too often), add --serial to get a more
```
skeggse, line 17: I haven't been seeing ava hang - do you have any repro steps? What's your system configuration?

tests.sh
```diff
@@ -0,0 +1,27 @@
...
+# deterministic record that will log up to the test prior to the one that hangs... But, as
+# is often the case with race conditions, the instrumentation may eliminate the problem
+# you're trying to track down...
+ava --no-color --timeout 30000 --verbose
```
skeggse, line 21: I wouldn't mind adding a `--timeout` to the default `test` command.

tests.sh
```diff
@@ -0,0 +1,27 @@
...
+eslint .
+prettier --check '**/*.(html|json|md|sublime-project|ts|yml)'
+nyc ava
```
skeggse, line 24: This will re-run ava - what do you think about combining this with the initial `ava` invocation?

Dockerfile-tests
```diff
@@ -0,0 +1,46 @@
...
+  && echo -n "apk " && apk --version \
+  && apk update \
+  && apk add --no-cache bash \
+  && rm -r /usr/share/man \
+  && true
+
+FROM beequeue-base as beequeue-node_modules
+
+# Ensure that tools for npm are available (for node-gyp in particular)
+# to build node_modules
```
skeggse, line 20: Are there any dependencies of bee-queue that actually require this?

Dockerfile-tests
```diff
@@ -0,0 +1,46 @@
...
+
+WORKDIR /bee-queue
+
+COPY --from=beequeue-node_modules /bee-queue/node_modules ./node_modules/
+
+# We use a script for Docker-based testing (instead of npm test) so
+# that the dev cycle doesn't have to wait for the npm install
+# triggered by editing e.g. the scripts stanza in package.json
+ENV TEST_SCRIPT $TEST_SCRIPT
+CMD ./${TEST_SCRIPT}
```
skeggse, line 43: It seems to me like this is trivially overridden when running docker, but maybe it's a docker-compose limitation? I'd be in favor of hardcoding the test script for the `CMD`.

Dockerfile-tests
```diff
@@ -0,0 +1,46 @@
...
+
+# Now copy all the source files
+COPY . ./
```
skeggse, line 46: This will copy in the `node_modules` and some other unnecessary things like the md files etc, right? I've often taken the approach of using the `.dockerignore` as an allowlist and only including the things that are important.

Dockerfile-tests
```diff
@@ -0,0 +1,46 @@
...
+
+FROM beequeue-base as beequeue-node_modules
+
+# Ensure that tools for npm are available (for node-gyp in particular)
+# to build node_modules
+RUN true \
+  && apk add --no-cache python make g++ \
+  && true
+
+WORKDIR /bee-queue
```
skeggse, line 25: Maybe this is a convention that I'm not familiar with, but it strikes me as really weird to introduce new root directories. Would we put this somewhere else like `/opt` or something?

tests.sh
hughsw: Heh, this PR captures my system config. ;-)  It's Docker for Mac.  Within this framework it's intermittent, and less likely when the `--serial` option is given to `ava`.   There are perhaps three tests that sometimes fail.  But, given ava's relative secrecy about what's being run, it's hard to figure out which test is hung.  I'm new to a lot of the tools you're using, so I'm a bit tentative about my conclusions.

 If I run Node from the command-line on the Mac (using the brew-installed Node 10), the tests almost never succeed..., that's what got me working on Docker and Bee-Queue.

Dockerfile-tests
hughsw: Or something... ;-)   Seriously, I work with literally dozens of different Docker containers that have a complex, Docker-based CICD system.  Each container in this universe typically does just one thing (Express server, HTML->PDF renderer, bee-queue workers).  Long ago we tried following file-system guidelines, but it just meant that bits and pieces of the one thing were spread out, and were hard to find, and because of the "or something" factor, not predictable across developers.  The pattern that has emerged on the projects I work on on is to put everything in a top-level directory.  That way, when you have to connect to the container and poke around (via some ssh chain), everything that matters is in that one place.

Dockerfile-tests
hughsw: `node-gyp` is the bad boy here.  Has been a PITA for years.  Something in the set of bee-queue (dev?) packages needs it.  We could get more nuanced if necessary, but this is a simple solution right now.

Dockerfile-tests
hughsw: We're trying to avoid having `npm ci` get run each time we tweak a test.  Docker is clever about caching and not redoing work it can prove it already has the results of, but it doesn't parse files for semantics.  So, if `package.json` is altered in any way from prior builds, the `npm ci` step gets re-run.  By relying on a separate script, that's copied into the image file system after the `npm ci` step,  that step is not re-run and the dev cycle is much quicker.  Having a script also supports running a custom script and other dev-type work -- nothing says you have to run tests in the script....

I'm approaching this the way I work in other projects -- I'm new to the Bee-Queue culture and if/how you want to enable container-based development and use of Bee-Queue.

Dockerfile-tests
hughsw: Originally it was hard-wired, but that meant any new dev work required changing the one-true test script.  This way, the one-true script is the default, but dev can use any script.

Dockerfile-tests
skeggse: :shrug: works for me!

Dockerfile-tests
skeggse: What I mean to say is that (as I understand it) when you use `docker run` you can just pass additional arguments to override the `CMD`.

Dockerfile-tests
skeggse: Strange - that isn't happening for me. It's happening for you on alpine or just on mac?

tests.sh
skeggse: Yeah ava was an early adoption and while on the one hand I'd kinda prefer jest for its stability and support I do still like the raw speed that ava provides. If only the author of `ava-spec` didn't deprecate it in favor of Jest...

I might have to try and get my hands on a mac to repro what you're seeing - it'd be nice if these tests succeeded, and race conditions like the ones you're talking about concern me.

Dockerfile-tests
skeggse: Thanks for sharing - it's interesting to hear how folks use Docker in development scenarios. I've only really used it for shipping production code in deployment scenarios. I'm cool with this, though I feel like the tweaking of the `scripts` object is rare enough that this is somewhat of a weird workflow optimization.

Dockerfile-tests
hughsw: I was a bit terse yesterday, long day.  It's somewhat complicated.

`nody-gyp` is used by npm when C/C++ libraries need to be compiled for npm modules.  I've seen it used for packages that need to wrangle bits and bytes to construct and/or unpack binary C structs to send over the wire, e.g. Sequelize to talk to SQLservers.  I'm guessing that the `redis` library wants to do the same thing to talk to Redis -- just guessing.

AFAICT, npm has several options about how to proceed in a given `npm install` environment:
- pull in pre-compiled libraries for the platform
- pull in node-gyp tools and compile the C/C++ code 
- fall back to a Javascript implementation
 
Using Alpine brings lots of these issues to the fore.  It can be a bit tedious, but it does help shed light on some subtle background matters.  Using Debian or Ubuntu won't have this friction...

If you look at the Hello World example PR you'll see that I don't pull in the build system, and the example works.  If you remove the `--quiet` option you'll see that there are numerous errors emitted by npm regarding not finding Python and that its "not OK".  Yet, the install succeeds and the example works.

If I add `apk add` Python, the build then emits errors about `make` not being present, yet the example works.  And if I add `make` then it emits errors about not having `g++`, etc.

I'm guessing that in the face of these errors, the build for `redis` falls back to a native JS implementation -- again, just guessing.

For the Hello World example this underlying turmoil is irrelevant, and whatever npm has installed is good enough.

But, for starting on a Docker configuration that wants to be reliable vis-a-vis Bee-Queue testing and performance, I want to use a robust build chain.  So, that's what's in this PR.  In the long run (if there is one) we would certainly want to have the testing vet both versions of the `redis` runtime (as Travis does with various Node versions).  I would also want to build/test with a Debian base image for similar exposure to different underlying tool chains.

hughsw: And, I'm stubborn about continuing to use a frictional build chain if it helps us isolate and correct race conditions that are not manifest in the standard CI chain.

Dockerfile-tests
hughsw: Yes, but it's a bit harder in the `docker-compose` setting.  And these are just initial versions -- I really don't have a good sense if anyone besides me is interested in this Dockerization effort.

I have a PR in the works that provides numerous cli arguments on the `go.sh` script which get communicated into the containers via env vars through `docker-compose`.

Dockerfile-tests
hughsw: The `.dockerignore` file has these three entries:
```
/.git
/.nyc_output
/node_modules
```

Yes, we copy everything else.  We could get more nuanced, but without guidelines and/or evidence for what not to copy, I don't see a reason to not include the whole package.  One change I have made in other projects is to use `git ls-files` to construct the list of files to copy, but that gets complex, something to avoid early on.

tests.sh
hughsw: We should discuss the purpose of this or other scripts we include.  As it stands, this script supports a dev-cycle (my dev cycle) whereby you get stuff working, you mollify the styling police, you move towards deployable coverage, etc.  I've gone back and forth on just doing `npm test`, but it fails too often and the output isn't that helpful when you start on forensic work.

tests.sh
hughsw: I will spend a little time on looking at this PR/branch in a Docker system on an underlying Linux kernel (unlike Mac, which uses the moby kernel emulator or whatever you call it).

hughsw: Experiment:

I used the PR branch on a Digital Ocean droplet which has an Ubuntu-based Docker environment -- So, stable, cloud-centric Linux kernel and mainstream Ubuntu/Debian Docker.

I used this script, `npmtest.sh`:
```bash
#!/bin/sh                                                                                                                                                                                                                                

exec npm test
```
Run via:
```
(while ./docker-tests.sh npmtest.sh ; do echo && echo timestamp && date && echo ; done) > /tmp/beequeuetests.log 2>&1
```
On this Droplet a new "timestamp" line would appear in the log file every 22 seconds or so, until... either:
- the while loop to stopped, meaning an explicit test failure
- the number of "timestamp" lines in the log file  stopped increasing, meaning a hung test

Out of 75 execs of `npm test` I got:
- 67 successful runs
- 7 hugh tests
- 1 failure

So, overall 89% success rate.

The one failure (which I have occasionally seen on Mac):
```
bee-queue_1  |   139 passed
bee-queue_1  |   1 failed
bee-queue_1  |
bee-queue_1  |   helpers-test › finallyRejectsWithInitial waits for the returned Promise
bee-queue_1  |
bee-queue_1  |   /bee-queue/test/helpers-test.js:85
bee-queue_1  |
bee-queue_1  |    84:     );
bee-queue_1  |    85:     t.true(measure() >= 10);
bee-queue_1  |    86:     t.true(stub.calledOnce);
bee-queue_1  |
bee-queue_1  |   Value is not `true`:
bee-queue_1  |
bee-queue_1  |   false
bee-queue_1  |
bee-queue_1  |   measure() >= 10
bee-queue_1  |   => false
bee-queue_1  |
bee-queue_1  |   measure()
bee-queue_1  |   => 7.332423
bee-queue_1  | npm ERR! Test failed.  See above for more details.
bee-queue_bee-queue_1 exited with code 1
```
The 7 hung runs had all completed 139 of the tests.   I'll have to work a little to figure out which test(s) were the one(s) that hung.

hughsw: I have goodish news.  So far, using my uuid-based approach (#237) to fixing the Job.id race condition, I am not seeing any of the non-deterministic test failures and hangs.

hughsw:
This PR is made obsolete by #239 
I will keep this as a draft until the Docker-based setting development feature is stable.

Dockerfile-tests
skeggse: >I'm guessing that the redis library wants to do the same thing to talk to Redis -- just guessing.

That was once the case - redis used a library called `hiredis` - but that hasn't been the case for a long time.

I think a general (minor) goal of this project is to not have any C/C++ dependencies, but I don't know that that needs to block this PR.

Dockerfile-tests
skeggse: :shrug: ok, thanks - haven't really used `docker-compose` much.

Dockerfile-tests
skeggse: >As an aside, Dockerfile `COPY` is a PITA when it comes to handling directory trees.

My limited exposure to Docker has taught me this much :see_no_evil:

tests.sh
skeggse: I'm specifically wondering about the inclusion of both `ava` and `nyc ava` in this script - not the use of a separate `tests.sh` file. Not sure if that was apparent from my original comment.

PR #236, opened 2020-04-17
docs(examples): Add Express example using Docker containers

hughsw:
Dockerized version of the Express example.

Me wonders if the intent was to add the two parameters of the API call as numbers, rather than catenate them...?

### DIFF
--- a/examples/docker/02-express-1/Dockerfile
+++ b/examples/docker/02-express-1/Dockerfile
@@ -1,0 +1,10 @@
+FROM node:lts-alpine
+
+# Minimalist image with bee-queue package and local JS files
+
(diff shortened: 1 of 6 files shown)


compwright:
@hughsw could you rebase this? the CI checks are stuck and that will un-stick them.

compwright:
@hughsw could you rebase this please? This will re-trigger the CI checks.

compwright:
Thank you for this submission. However, merging this will introduce duplicate example code that someone (me) will have to maintain down the road. Since I haven't seen anyone asking for these, and since generally examples are for reviewing, not necessarily for running, I think I'm going to forego merging this.

PR #237, opened 2020-04-18
WIP/RFC -- Fix dropped jobs UUID

hughsw:
Alternate fix for #78 and #189 
Builds on #193 , returning to its original intent

Uses random hex string for Job.id, generated and registered in Queue prior to adding job to Redis.


hughsw:
This approach breaks the semantic that custom job id's don't affect `checkHealth().newestJob`, so the `Queue Health Check should not report the latest job for custom job ids` fails.

The semantic could be implemented, but:
- it's additional complexity to support an obscure semantic
- I don't know how much the existing semantic is relied upon, it takes some hoop-jumping (as per the test) to observe -- what is the justification for custom ids being treated differently in terms of `newestJob` ?

hughsw:
Actually, it's not hard to preserve the existing semantic, but I'd like some discussion about the necessity of this semantic and the test.

hughsw:
Wrapping `Job.fromId` in order to test removal from 'active' prior to processing is problematic:  This approach replaces a global resource with a one-shot modification.  In a parallel testing framework, there's no guarantee that the test that replaces `fromId` is the one that uses the one-shot replacement...

hughsw:
So the one remaining test failure is baffling to me.  The deepEqual of two Set object fails because the items in the set are in different orders (even though the set of items is the same).  It is reasonable that the code changes for uuid Job.id would cause the orders to be different, but my understanding is that Set should be deepEqual regardless of the order.  Any insight into the failure?

If I sort the error arrays (and compare the arrays rather than go to the trouble of making sets), the test passes, so that's likely what I'm going to add to this PR.

skeggse:
I'll try and take a look later tomorrow. I'm also curious what you think of #232.

lib/helpers.js
```diff
@@ -24,6 +24,11 @@ function finallyRejectsWithInitial(promise, fn) {
 
 const promiseUtils = require('promise-callbacks');
 
+//  A simple way to get up to 14 random hex digits, not cryptographic
+const randomHex14 = () => {
+  return Math.floor(Math.random() * Number.MAX_SAFE_INTEGER).toString(16);
```
hughsw, line 29: We could prefix this string with the value of `Date.now()` so as to have IDs be ordered, if we believe this undocumented feature of existing IDs should be supported.  E.g. `1587387486767-10ef4cf53faaf`

skeggse: It's specifically problematic for us because we use it to compute influx metrics - every ten seconds we pull `checkHealth` and subtract the previous `id` from the new value.

hughsw: Hmmm, this PR doesn't affect the growth of `newestJob` -- specifically that semantic is preserved.  So, subtracting values and dividing by elapsed time will still give a throughput metric.

What is changed is that `newestJob` it is no longer an ID that's been in the system;  it is only there to be the counter (so arguably the name is wrong).  We could add a correct name, `jobCount` and deprecate the `newestJob` name...

I was talking about changing the ID to have a timestamp prefix solely to ensure that the IDs are lexically monotonic.

skeggse: >Hmmm, this PR doesn't affect the growth of `newestJob` -- specifically that semantic is preserved.

I'm not sure I understand how this works, then. The timestamp prefix wouldn't let us compute the delta between two `id`s to get a throughput metric - it would result in the amount of time elapsed instead of the number of jobs created - right?

>And, can we revisit why non-default-IDs (client IDs) are excluded from the count...?

Yes, that'd be fantastic! Not by design, just never prioritized.

hughsw: I think I understand.  

I had assumed that your throughput was based on taking two `checkHealth().newestJob` measures -- their difference is the number of jobs delta, and the difference in `Date.now()` between those two measures gets the time difference-- that's how I'm computing job throughput, as per hints in README.  Any Queue instance can make such a throughput determination and it doesn't need to handle any jobs to do so.

You are using the difference between two specific `job.id` to get the jobs delta, and also using difference between `Date.now()` for the time difference.  You have sentinel jobs (or something like that) and they are the ones that go into making the jobs-count delta.  Do understand correctly about using `job.id` and about specific jobs being used for throughput?

My discussion about a prefix on the `job.id` was solely an attempt to preserve the semantic that the existing `job.id` strings have a lexical ordering by virtue of coming from the incrementing counter.

hughsw: My bias, based on various experiences, is to avoid putting any semantics on identifier values as it's one form of lava that is particularly hard to break free of...

compwright, line 29: @hughsw as you noted, Math.random() is not cryptographically secure. Thus this implementation will be prone to auto-generated job IDs colliding when run at sufficient scale.

Could you revise using [crypto.randomBytes()](https://nodejs.org/docs/latest-v12.x/api/crypto.html#crypto_crypto_randombytes_size_callback) or equivalent instead?

compwright (changes requested):

compwright:
@hughsw this also will need a rebase.

compwright (approved):
Approved pending rebase

lib/lua/addDelayedJob.lua
```diff
@@ -1,5 +1,5 @@
 --[[
-key 1 -> bq:name:id (job ID counter)
+key 1 -> bq:name:id (job counter)
```
bradvogel, line 2: What is stored in `bq:name:id`? My understanding from this PR is that it's the number of jobs that don't have a custom id? What's that used for?

compwright: I believe that is correct, it's poorly named, and appears to come from the fact that auto-generated IDs are currently sequential. So job ID 1000 would be the 1000th job. This PR of course changes that but does not rename to try to maximize compatibility, replacing the stat with the actual count instead of the actual last job ID from which you would infer the count.

bradvogel: Where is that information ("the number of jobs that don't have a custom id") used?

PR #239, opened 2020-04-20
feat(tests): Add docker-test.sh for running tests in a Docker container

hughsw:
Use a single script, `docker-test.sh` to build an image and run a container to do testing.

This is a less intrusive, more focussed version of #234 

```
$ ./docker-test.sh  --help

docker-test.sh -- Run standard bee-queue tests in a Docker container with a Redis server

Usage: docker-test.sh [command ...args]

The optional command and args specify what to run in the container (default: npm test).

You can use tools or run scripts from package.json, e.g.

  ./docker-test.sh  npx ava --serial --fail-fast --verbose --no-color --timeout 30000
  ./docker-test.sh  npm run coverage

If you want to work interactively in the container, run a shell, e.g.

  ./docker-test.sh  bash

Notes:
- The very first time this script is used it takes a while for Docker to
  download the base image and to populate its caches.
- You can ignore the Transparent Huge Pages warning from the Redis server.
```


docker-test.sh
```diff
@@ -0,0 +1,92 @@
...
+
+# Strict, and loud failure
+set -euo pipefail
+trap 'rc=$?;set +ex;if [[ $rc -ne 0 ]];then trap - ERR EXIT;echo 1>&2;echo "*** fail *** : code $rc : $DIR/$SCRIPT $ARGS" 1>&2;echo 1>&2;exit $rc;fi' ERR EXIT
+ARGS="$*"
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+SCRIPT="$(basename "${BASH_SOURCE[0]}")"
+
+
+if [ "${1:-}" = -h -o "${1:-}" = --help ] ; then
```
skeggse, line 11: Wow I'm having a real hard time parsing this. I'm guessing that bash is interpreting `-o` as an OR. Would this be equivalent?
```suggestion
if [ "${1:-}" = -h ] || [ "${1:-}" = --help ] ; then
```

docker-test.sh
```diff
@@ -0,0 +1,92 @@
...
+eval "$@"
+
+EOF
+chmod +x entrypoint.sh
+
+# Make a minimal version of package.json to avoid unnecessary `npm ci` when rebuilding the image
+node > package-min.json <<EOF
+const {dependencies, devDependencies} = require('./package.json');
+console.log(JSON.stringify({dependencies, devDependencies}));
+EOF
```
skeggse, line 67: `node` supports a `-p` option for doing some of this:
```suggestion
node -p 'const {dependencies, devDependencies} = require("./package.json");
JSON.stringify({dependencies, devDependencies});'
```
not sure that it's any nicer, though...

docker-test.sh
hughsw: My old-school colors are showing! ;-) I never keep all this straight between what bash does internally and when it uses the `[` command (`/bin/test`).  I prefer the direction you're moving it -- it may need double square brackets.  I'll clean it up.

hughsw: Done, with extras that bash `[[` needs in order to do simple matching.

docker-test.sh
hughsw: It could be made to work...  but why break the here-doc pattern of this script?

compwright:
@hughsw would you mind giving this a rebase?

compwright (approved):

compwright:
Resubmitted as https://github.com/bee-queue/bee-queue/pull/554

PR #252, opened 2020-05-12
update type declaration

cjroebuck:
redis option should be either options object or instance


cjroebuck:
Also added status field to job type

skeggse (approved):

skeggse:
Thanks!

Issue #253, opened 2020-05-19
Can't use rediss TLS url - upgrade node-redis version


b-jam:
To use redis with TLS - i.e. `rediss://` the node-redis version must be ideally greater than 3. 

The actual version implementing `rediss` was 2.8.X, so at a minimum that version which contains this commit.
https://github.com/NodeRedis/node-redis/commit/5d6e471d0c14df39974000427719c5d85fa4f21f

See https://github.com/NodeRedis/node-redis/issues/1331

b-jam:
The current version in bee-queue is 2.7.x

skeggse:
I'm gonna say we're not gonna address this before v2, which will use `ioredis`. Maybe if npm supported conditional dependencies or something but that'd be lunacy. The workaround described in #257 _should_ work but I haven't investigated.

PR #257, opened 2020-05-31
Upgrade redis to 3.0.2

b-jam:
Fixes https://github.com/bee-queue/bee-queue/issues/253

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2249,6 +2249,11 @@
       "integrity": "sha1-3zrhmayt+31ECqrgsp4icrJOxhk=",
       "dev": true
     },
+    "denque": {
+      "version": "1.4.1",
+      "resolved": "https://registry.npmjs.org/denque/-/denque-1.4.1.tgz",
+      "integrity": "sha512-OfzPuSZKGcgr96rf1oODnfjqBFmr1DVoc/TrItj3Ohe0Ah1C5WX5Baquw/9U9KovnQ88EqmJbD66rKYUQYN1tQ=="
+    },
     "detect-indent": {
       "version": "4.0.0",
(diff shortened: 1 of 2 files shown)


b-jam:
Unsure why CI is failing, but it works for me in an app using TLS, and `yarn test` runs fine locally

skeggse:
This project uses `npm` - please use `npm` to upgrade `redis` as that'll avoid the `yarn.lock` and will update the `package-lock.json`. CI is failing because the two package files aren't in sync.

b-jam:
Ah. It looks better now, but incompatible with node 4 so that checks failing still

b-jam:
Hey @skeggse is it important to keep node 4 happy, or can I perhaps remove it from travis, and add a lts like `lts/erbium` ?

skeggse:
We'll have to wait to deprecate Node 4 support for bee-queue v2, which is currently on hold pending an owner.

b-jam:
Sure, can I merge this into `2.0dev` then instead ?

b-jam:
Nevermind - I see thats using ioredis anyway

skeggse:
One of 2.0's main features is moving to `ioredis`, so I'm not sure that's the right move. Did Redis 3 drop Node 4 support?

If this is something you want internally, you should be able to just pass a Redis 3 connection in to the Queue constructor.

b-jam:
https://github.com/NodeRedis/node-redis/blob/master/.github/workflows/tests.yml
looks like it judging from their CI which only runs v6+

ah that sounds like a better solution than forking. Will check it out

b-jam:
As a note, I tried using a 3.0 connection but it just didn't work. Didn't do a deep dive why, just that it didn't work after I changed it to a 3.0 connection to the same redis, TLS or not.

I will patiently wait for bq 2.0 :)

PR #275, opened 2020-08-11
chore(deps-dev): update sinon to v8

skeggse:
In lieu of #274 because it's blocked on dropping Node 8 support.


shils:
@skeggse is there a timeline for dropping Node 4 support?

PR #290, opened 2020-10-07
Update Redis configuration type

danieldunderfelt:
The Redis QueueSettings prop can accept a node_redis client object, but this wasn't reflected in the typings. If a Typescript user tried to set a RedisClient, Typescript would not compile without using `@ts-ignore`.

This PR fixes that by adding the `RedisClient` type to the `redis` prop as a union type.


skeggse (approved):

skeggse:
Thanks!

PR #296, opened 2020-10-23
Clear jobs if we try to fetch them and they aren't there

tmeasday:
In certain pathological cases, job data can get lost (to be honest not totally sure how). It seems sensible to just drop the job if we find this, rather than the current behaviour which is to constantly send the job through a `waiting => active => waiting` loop.

### DIFF
--- a/README.md
+++ b/README.md
@@ -349,6 +349,7 @@ The `settings` fields are:
 - `removeOnFailure`: boolean. Enable to have this worker automatically remove its failed jobs from Redis, so as to keep memory usage down. This will not remove jobs that are set to retry unless they fail all their retries.
 - `quitCommandClient`: boolean. Whether to `QUIT` the redis command client (the client it sends normal operations over) when `Queue#close` is called. This defaults to `true` for normal usage, and `false` if an existing `RedisClient` object was provided to the `redis` option.
 - `redisScanCount`: number. For setting the value of the `SSCAN` Redis command used in `Queue#getJobs` for succeeded and failed job types.
+- `clearMissingJobs`: boolean. Enable to clear out job references if data is missing.
 
 ### Properties
 
--- a/lib/job.js
+++ b/lib/job.js
@@ -28,7 +28,9 @@ class Job extends Emitter {
       .then((data) =>
         data
           ? Job.fromData(queue, jobId, data)
-          : queue.removeJob(jobId).then(() => null)
+          : queue.settings.clearMissingJobs
+          ? queue.removeJob(jobId).then(() => null)
+          : null
       );
 
     if (cb) helpers.asCallback(promise, cb);
--- a/lib/job.js
+++ b/lib/job.js
@@ -25,7 +25,11 @@ class Job extends Emitter {
           client.hget(queue.toKey('jobs'), jobId, done)
         )
       )
-      .then((data) => (data ? Job.fromData(queue, jobId, data) : null));
+      .then((data) =>
+        data
+          ? Job.fromData(queue, jobId, data)
+          : queue.removeJob(jobId).then(() => null)
+      );
 
     if (cb) helpers.asCallback(promise, cb);
     return promise;
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -676,18 +676,18 @@ class Queue extends Emitter {
             return;
           }
 
-          this.running += 1;
-          this.queued -= 1;
-          if (this.running + this.queued < this.concurrency) {
-            this.queued += 1;
-            setImmediate(jobTick);
-          }
-
           if (!job) {
             // Per comment in Queue#_waitForJob, this branch is possible when
             // the job is removed before processing can take place, but after
             // being initially acquired.
             return;
+          }
+
+          this.running += 1;
+          this.queued -= 1;
+          if (this.running + this.queued < this.concurrency) {
+            this.queued += 1;
+            setImmediate(jobTick);
           }
 
           return this._runJob(job).then((results) => {
--- a/test/queue-test.js
+++ b/test/queue-test.js
@@ -1440,7 +1440,7 @@ describe('Queue', (it) => {
     });
 
     it('clears a waiting job that is missing data', async (t) => {
-      const queue = t.context.makeQueue();
+      const queue = t.context.makeQueue({clearMissingJobs: true});
 
       const job = await queue.createJob({foo: 'first'}).save();
 
--- a/test/queue-test.js
+++ b/test/queue-test.js
@@ -1445,8 +1445,8 @@ describe('Queue', (it) => {
       const job = await queue.createJob({foo: 'first'}).save();
 
       // Delete the job's data from redis
-      await new Promise((resolve) =>
-        queue.client.hdel(queue.toKey('jobs'), job.id, resolve)
+      await helpers.callAsync((done) =>
+        queue.client.hdel(queue.toKey('jobs'), job.id, done)
       );
 
       // Create a second job
@@ -1466,12 +1466,10 @@ describe('Queue', (it) => {
 
       // By default the job id would have been moved to the active list (soon to stall)
       // So we just check it is no longer there
-      await new Promise((resolve) =>
-        queue.client.llen(queue.toKey('active'), (_, active) => {
-          t.is(active, 0);
-          resolve();
-        })
+      const active = await helpers.callAsync((done) =>
+        queue.client.llen(queue.toKey('active'), done)
       );
+      t.is(active, 0);
     });
 
     it('processes a job that auto-retries', async (t) => {
--- a/test/queue-test.js
+++ b/test/queue-test.js
@@ -1439,6 +1439,41 @@ describe('Queue', (it) => {
       t.is(err.message, `Job ${job.id} timed out (10 ms)`);
     });
 
+    it('clears a waiting job that is missing data', async (t) => {
+      const queue = t.context.makeQueue();
+
+      const job = await queue.createJob({foo: 'first'}).save();
+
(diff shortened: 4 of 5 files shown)


lib/queue.js
```diff
@@ -673,20 +673,20 @@ class Queue extends Emitter {
...
             return;
           }
 
+          this.running += 1;
+          this.queued -= 1;
+          if (this.running + this.queued < this.concurrency) {
+            this.queued += 1;
+            setImmediate(jobTick);
+          }
+
```
skeggse, line 689: What's the intent behind moving these lines? Seems like we're interfering with the intended function of perpetuating one logical "thread"/worker to read and process jobs by doing this.

test/queue-test.js
```diff
@@ -1297,6 +1297,41 @@ describe('Queue', (it) => {
...
 
+    it('clears a waiting job that is missing data', async (t) => {
+      const queue = t.context.makeQueue();
+
+      const job = await queue.createJob({foo: 'first'}).save();
+
+      // Delete the job's data from redis
+      await new Promise((resolve) =>
+        queue.client.hdel(queue.toKey('jobs'), job.id, resolve)
+      );
```
skeggse, line 1308: As a general rule, I'd like to handle errors (yes, even in tests - just in case something strange breaks). Probably easiest to just use `callAsync` from `promise-callbacks` (this'll need an additional import clause).
```suggestion
      await callAsync((done) => queue.client.hdel(queue.toKey('jobs'), job.id, done));
```

test/queue-test.js
```diff
@@ -1297,6 +1297,41 @@ describe('Queue', (it) => {
...
+      t.is(callCount, 1);
+
+      // By default the job id would have been moved to the active list (soon to stall)
+      // So we just check it is no longer there
+      await new Promise((resolve) =>
+        queue.client.llen(queue.toKey('active'), (_, active) => {
+          t.is(active, 0);
+          resolve();
+        })
+      );
```
skeggse, line 1332:
```suggestion
      const active = await callAsync((done) => queue.client.llen(queue.toKey('active'), done));
      t.is(active, 0);
```

lib/job.js
```diff
@@ -27,7 +27,11 @@ class Job extends Emitter {
           client.hget(queue.toKey('jobs'), jobId, done)
         )
       )
-      .then((data) => (data ? Job.fromData(queue, jobId, data) : null));
+      .then((data) =>
+        data
+          ? Job.fromData(queue, jobId, data)
+          : queue.removeJob(jobId).then(() => null)
+      );
```
skeggse, line 34: This feels vaguely dangerous. Could we make it opt-in? What do you think?

tmeasday: Sure, makes sense!

lib/queue.js
tmeasday: The previous code was (or at least seems) logically wrong -- when `job` is falsey, we immediately stop this worker and never decrement `this.running` (which happens on line 691 below), so it grows without bound.

As it happens we only ever check the sum `this.running + this.queued` so the combination of decrementing one and incrementing the other actually has no practical effect anyway, which I guess is the reason why this "bug" is in the released code.

tmeasday: I guess there is also a control flow change inherent in moving this code block. (the `setImmediate(jobTick)`). 

We are about to exit this promise anyway, which will call `setImmediate(jobTick)` in the `finally` block. Assuming the next job we call isn't missing too, we'll start queuing up new workers if there's concurrency available so it isn't a big deal, but if you are concerned we could just do the reincrement/redecrement in the `if (!job)` instead.

lib/job.js
tmeasday: I added the option in https://github.com/bee-queue/bee-queue/pull/296/commits/3671e4d74ad276b78d3c8a103955a8c37756e207

However, thinking about it some more this feels like a mistake, as:

 1. It isn't turned on by default so isn't as well exercised in other tests etc.
 2. If the missing job thing happens (admittedly I don't fully understand *why* it happened to us) the alternative is to have a job pass through the queue ad-infinitum. I'm not sure that's what anyone would ever want?

Just as an FYI we've been running this code in production for a little while and it hasn't caused us any problems (I'm not sure if any data has gone missing in that time, though).

lib/queue.js
skeggse: urgh I hate how this code is structured. every time I read it I think I spot something wrong only to figure out I'm mistaken. if we add a test that fails before this PR and succeeds after it regarding spoolup then I'm cool with it :)

lib/job.js
```diff
@@ -27,7 +27,13 @@ class Job extends Emitter {
           client.hget(queue.toKey('jobs'), jobId, done)
         )
       )
-      .then((data) => (data ? Job.fromData(queue, jobId, data) : null));
+      .then((data) =>
+        data
+          ? Job.fromData(queue, jobId, data)
+          : queue.settings.clearMissingJobs
```
skeggse, line 33: I could also imagine some pluggable handler for this case, but that's out of scope.

README.md
```diff
@@ -335,6 +335,7 @@ The `settings` fields are:
 - `removeOnFailure`: boolean. Enable to have this worker automatically remove its failed jobs from Redis, so as to keep memory usage down. This will not remove jobs that are set to retry unless they fail all their retries.
 - `quitCommandClient`: boolean. Whether to `QUIT` the redis command client (the client it sends normal operations over) when `Queue#close` is called. This defaults to `true` for normal usage, and `false` if an existing `RedisClient` object was provided to the `redis` option.
 - `redisScanCount`: number. For setting the value of the `SSCAN` Redis command used in `Queue#getJobs` for succeeded and failed job types.
+- `clearMissingJobs`: boolean. Enable to clear out job references if data is missing.
```
skeggse, line 338: Thanks!

lib/job.js
skeggse: Hmm on further consideration I do want to understand how we get into this state. Maybe we reach `_finishJob` and the `multi`-`exec` doesn't remove the job from `waiting` but it's actually in there? I'm not able to focus enough to really reason about that scenario. I wonder about whether we should be adopting safety belts like this (reasonable) or whether we should avoid them and strive for end-to-end implementation correctness (pipe dream?) as a general rule for bee-queue.

lib/queue.js
tmeasday: Any hints on how to write this test? Should I create a bunch of jobs, remove the data for one (manually via a redis call) and then start the queue w/ some concurrency?

skeggse: yeah that seems like a good start. I'd also block all jobs in some appropriate manner and wait for the appropriate number to start to make sure the correct number of worker "threads" run.

lib/job.js
tmeasday: I kind of share you unease here. Maybe this whole PR is the wrong approach.

I was thinking, is it possible this comes out as a race condition between the `checkStalledJobs` script and the `_finishJob` `multi` transaction?

1. `checkStalledJobs` will move the job from `active` => `waiting` if it is stalling

2. `_finishJob` removes the job from `active` and `stalling` but not `waiting` (or `delayed` for that matter). 

I'm not quite sure what the scenario would be, I guess maybe if the machine handling the job is under heavy load it might stall, but still eventually finish the job just as another machine is moving the job to `waiting`?

In which case the solution would be simply to ensure that `_finishJob` cleans up `waiting` too.

WDYT? I'm not really sure how the issue came about but it was definitely associated with machines that were struggling and a lot of jobs stalling (see other PR).

skeggse: > In which case the solution would be simply to ensure that `_finishJob` cleans up `waiting` too.

Yeah this is where my thinking led me. I'm pretty hesitant to delete data in case there's an implementation error in `bee-queue` - we wouldn't want to outright drop data that'll never get completed. Ideally, of course, there isn't an implementation error in `bee-queue` :p

I think I'm in favor of closing the gap in `_finishJob`, though I haven't thought through the performance implications fully. Maybe that's the next step before we start deleting data? Whereas the `active` key is bounded by the number of workers, the `waiting` key is unbounded; we'd incur a worst-case `O(N)` cost for every job we complete. We can proooobably assume it'll be closer to the end of the list, and find/remove from that end - but that won't guarantee good performance. Maybe it's worth moving the `_finishJob` `MULTI` to a script so we can conditionally remove from `waiting` if it's no longer in `active`?

tmeasday: > I'm pretty hesitant to delete data

Point of clarification here--the data is already deleted, it is just the references to it in the `active` (et al) lists that remain. Not sure it changes much but just wanted to be clear.

> Maybe it's worth moving the _finishJob MULTI to a script so we can conditionally remove from waiting if it's no longer in active?

I can give that a go! Might be a few days before I get round to it.

tmeasday: > We can proooobably assume it'll be closer to the end of the list, and find/remove from that end - but that won't guarantee good performance.

I mean in almost all cases it won't be in there at all so that won't make much of a perf difference.

skeggse: > Point of clarification here--the data is already deleted, it is just the references to it in the `active` (et al) lists that remain. Not sure it changes much but just wanted to be clear.

Yeah good call-out - my mistake.

> I mean in almost all cases it won't be in there at all so that won't make much of a perf difference.

Using something like `LREM` will have the time complexity of `O(N)`, regardless of whether the value is in the list or not - that's the time cost I'm describing. The only way to eliminate that cost is by conditionally applying the `LREM`.

tmeasday: > The only way to eliminate that cost is by conditionally applying the LREM.

Yeah, sorry I was agreeing with you but I see now my comment was a bit ambigious.

PR #297, opened 2020-10-23
Ensure we don't call `unpack` with too many elements

tmeasday:
If there are many many stalled jobs, we can hit lua's built in stack limit when trying to unpack them all in the `checkStalledJobs` script. This does the unpacking in batches.

### DIFF
--- a/lib/lua/checkStalledJobs.lua
+++ b/lib/lua/checkStalledJobs.lua
@@ -12,6 +12,10 @@ if a jobId is not removed from the stalling set within a stallInterval window,
 we assume the job has stalled and should be reset (moved from active back to waiting)
 --]]
 
+-- It isn't clear what this value should be, it depends on memory available for the stack.
+-- This seems a reasonable limit.
+local maxUnpack = 1024
+
 -- try to update the stallBlock key
 if not redis.call("set", KEYS[1], "1", "PX", tonumber(ARGV[1]), "NX") then
   -- hasn't been long enough (stallInterval) since last check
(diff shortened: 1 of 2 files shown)


lib/lua/checkStalledJobs.lua
```diff
@@ -31,20 +35,24 @@ if next(stalling) ~= nil then
...
+  -- hog the job consumers and starve the whole system. not a great situation
+  -- to be in, but this is fairer.
+  local pushed = 0
   -- don't lpush zero jobs (the redis command will fail)
-  if #stalled > 0 then
-    -- lpush instead of rpush so that jobs which cause uncaught exceptions don't
-    -- hog the job consumers and starve the whole system. not a great situation
-    -- to be in, but this is fairer.
-    redis.call("lpush", KEYS[3], unpack(stalled))
+  while (pushed < #stalled) do
```
skeggse, line 44:
```suggestion
  while pushed < #stalled do
```

lib/lua/checkStalledJobs.lua
```diff
@@ -31,20 +35,24 @@ if next(stalling) ~= nil then
...
   end
   redis.call("del", KEYS[2])
 end
 
 -- copy currently active jobs into stalling set
-local actives = redis.call("lrange", KEYS[4], 0, -1)
-if next(actives) ~= nil then
-  redis.call("sadd", KEYS[2], unpack(actives))
+local actives, added = redis.call("lrange", KEYS[4], 0, -1), 0
+while (added < #actives) do
```
skeggse, line 53:
```suggestion
while added < #actives do
```

lib/lua/checkStalledJobs.lua
skeggse, line 44: The `#` length operator is re-computed on each reference - we should push this (and the later `#actives` condition) into a `local` to avoid this penalty.

skeggse (commented):
Thanks for this! Couple small tweaks, please :)

tmeasday:
@skeggse do you have any idea why linting started failing (for `README.md` on only two environments) when I commited your simple `lua` changes? (It seems like `lua` isn't linted anyway...?)

tmeasday:
Oh, I see there is a broken build on `master`. I think this is ready for review then (tests partially passing!)

skeggse (approved):

skeggse:
> Oh, I see there is a broken build on `master`. I think this is ready for review then (tests partially passing!)

Fixed now - please rebase!

tmeasday:
Done.

PR #315, opened 2020-11-09
Use a lua script to execute `finishJob`

tmeasday:
In doing so clear the key from waiting if it isn't in active. See #296

@skeggse -- I'm not sure if we need additional tests here or not. Was this what you had in mind?

### DIFF
--- a/lib/lua/finishJob.lua
+++ b/lib/lua/finishJob.lua
@@ -1,0 +1,47 @@
+
+--[[
+key 1 -> bq:name:active
+key 2 -> bq:name:waiting
+key 3 -> bq:name:stalling
+key 4 -> bq:name:failed
+key 5 -> bq:name:delayed
+key 6 -> bq:name:earlierDelayed
+key 7 -> bq:name:succeeded
+key 8 -> bq:name:events
+key 9 -> bq:name:jobs
+
+arg 1 -> job id
+arg 2 -> remove job?
+arg 3 -> job data
+arg 4 -> new job status
+arg 5 -> delayed time
+arg 6 -> job event data
+--]]
+
+-- If we don't find the job in the active list, ensure it hasn't been recently stalled
+if not redis.call("lrem", KEYS[1], 0, ARGV[1]) then
+  redis.call("lrem", KEYS[2], ARGV[1])
+end
+
+redis.call("srem", KEYS[3], ARGV[1])
+
+if ARGV[2] ~= "" then -- remove?
+  redis.call("hdel", KEYS[9], ARGV[1])
+else
+  redis.call("hset", KEYS[9], ARGV[1], ARGV[3])
+
+  if ARGV[4] == "failed" then
+    redis.call("sadd", KEYS[4], ARGV[1])
+  elseif ARGV[4] == "succeeded" then
+    redis.call("sadd", KEYS[7], ARGV[1])
+  elseif ARGV[5] ~= "" then
+    redis.call("zadd", KEYS[5], tonumber(ARGV[5]), ARGV[1])
+    redis.call("publish", KEYS[6], tonumber(ARGV[5]))
+  else
+    redis.call("lpush", KEYS[2], ARGV[1])
+  end
+end
+
+if ARGV[6] ~= "" then
+  redis.call("publish", KEYS[8], ARGV[6]);
+end
--- a/lib/lua/finishJob.lua
+++ b/lib/lua/finishJob.lua
@@ -11,11 +11,10 @@ key 8 -> bq:name:events
 key 9 -> bq:name:jobs
 
 arg 1 -> job id
-arg 2 -> remove job?
-arg 3 -> job data
-arg 4 -> new job status
-arg 5 -> delayed time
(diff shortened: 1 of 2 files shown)


lib/queue.js
```diff
@@ -588,45 +586,43 @@ class Queue extends Emitter {
...
+      this.toKey('earlierDelayed'),
+      this.toKey('succeeded'),
+      this.toKey('events'),
+      this.toKey('jobs'),
+      job.id,
+      removeJob ? 'true' : '',
+      job.toData(),
+      job.status,
+      delayedTime || '',
+      this.settings.sendEvents && JSON.stringify(jobEvent)
```
skeggse, line 622: does this work if we don't pass in an empty string?
```suggestion
      this.settings.sendEvents ? JSON.stringify(jobEvent) : ''
```

lib/queue.js
```diff
@@ -563,17 +563,15 @@ class Queue extends Emitter {
...
-      .srem(this.toKey('stalling'), job.id);
-
     const jobEvent = {
       id: job.id,
       event: status,
       data: err ? err.message : data,
     };
 
+    let removeJob = false;
+    let delayedTime;
```
skeggse, line 573: could move the default here
```suggestion
    let delayedTime = '';
```

lib/queue.js
```diff
@@ -588,45 +586,43 @@ class Queue extends Emitter {
...
+      this.toKey('stalling'),
+      this.toKey('failed'),
+      this.toKey('delayed'),
+      this.toKey('earlierDelayed'),
+      this.toKey('succeeded'),
+      this.toKey('events'),
+      this.toKey('jobs'),
+      job.id,
+      removeJob ? 'true' : '',
+      job.toData(),
```
skeggse, line 619: similar to reducing bandwidth requirements, we could unify these two parameters s.t. an empty string represents "remove job" and a non-empty string represents "don't remove job + update the data"

lib/lua/finishJob.lua
```diff
@@ -0,0 +1,47 @@
...
+arg 1 -> job id
+arg 2 -> remove job?
+arg 3 -> job data
+arg 4 -> new job status
+arg 5 -> delayed time
+arg 6 -> job event data
+--]]
+
+-- If we don't find the job in the active list, ensure it hasn't been recently stalled
+if not redis.call("lrem", KEYS[1], 0, ARGV[1]) then
```
skeggse, line 22: according to [the docs](https://www.lua.org/pil/3.3.html), lua considers `0` to be truthy - I think we need to use a comparison operator instead?

skeggse (commented):
>@skeggse -- I'm not sure if we need additional tests here or not. Was this what you had in mind?

This was indeed what I had in mind, thanks! If we don't already have a test that removes a job while it's processing (either artificially or by having multiple `Queue`s where another queue picks up the work and finishes it before another `Queue` that stalls finishes the work) and verifies that it doesn't get orphaned, I think we should add one.

It looks like all the tests are passing under CI. The CI errors are just commitlint, which I need to rethink given that we always squash PRs in `bee-queue`. If you're seeing some tests stall/hang locally let me know and I can look into it.

tmeasday:
>  If you're seeing some tests stall/hang locally let me know and I can look into it.

Oh, you saw my earlier message -- this was just because I had commented out the final publish part (as part of debugging) and had neglected to uncomment it. The tests have run fine since I actually ran the correct code ;)

lib/queue.js
tmeasday: I thought it was maybe clearer to have the default in the call.

lib/queue.js
tmeasday: :+1:

lib/lua/finishJob.lua
tmeasday: Ok, better make sure I add a test that proves this.

lib/queue.js
skeggse: sure - I'm impartial

lib/lua/finishJob.lua
tmeasday: @skeggse -- sorry it took me so long to get back to this. Any advice on how to achieve this? 

I tried the following but it doesn't seem to move the job into the `waiting` list (i.e the final assertion fails):

```js
      const queue = t.context.makeQueue({
        stallInterval: 1,
      });

      const job = queue.createJob({foo: 'bar'});
      await job.save();

      // Block the job from finishing
      const {done, next} = reef(1);
      queue.process(async () => {
        await done;
      });

      // Mark the jobs as stalling, so that Queue#process immediately detects
      // them as stalled.
      await forceStall(queue);
      await queue._doStalledJobCheck();

      const counts = await queue.checkHealth();
      t.is(counts.waiting, 1);
```

tmeasday:
@skeggse I just came back to this (wow what a delay!) and although the ticket is marked `awaiting-reply` I think I'm actually blocked on finishing it by this comment: https://github.com/bee-queue/bee-queue/pull/315#discussion_r594003991

Issue #147, opened 2018-09-21
bee-queue will not wait for redis re-connection and fails with error


denigmus:
Testing bee-queue for a busy application. After connection with redis-server is lost, bee-queue will not wait for re-connection, but will fail. Sure, re-connection is a task of a redis-client, but seems that bee can't process it correctly when client reports disconnection. I'm surprised, is that as per design or a bug? I didn't find any setting for re-connection attempts or/and interval for that. 

`A alertsQueue error happened: Redis connection to 127.0.0.1:6379 failed - read ECONNRESET object { Error: Redis connection to 127.0.0.1:6379 failed - read ECONNRESET
    at _errnoException (util.js:992:11)    at TCP.onread (net.js:618:25) code: 'ECONNRESET', errno: 'ECONNRESET', syscall: 'read' }`

denigmus:
I thought ioredis will improve the something ,but both bee-queue: 1.x and bee-queue 2.0dev behave the same on redis stop, here is from 2.0dev:


```
A alertsQueue error happened: connect ECONNREFUSED 127.0.0.1:6379 object { Error: connect ECONNREFUSED 127.0.0.1:6379
    at Object._errnoException (util.js:992:11)
    at _exceptionWithHostPort (util.js:1014:20)
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1186:14)
  code: 'ECONNREFUSED',
  errno: 'ECONNREFUSED',
  syscall: 'connect',
  address: '127.0.0.1',
  port: 6379 }
(node:14928) UnhandledPromiseRejectionWarning: ReplyError: NOSCRIPT No matching script. Please use EVAL.
    at parseError (pathhere\node_modules\ioredis\node_modules\redis-parser\lib\parser.js:179:12)
    at parseType (pathhere\node_modules\ioredis\node_modules\redis-parser\lib\parser.js:302:14)
(node:14928) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled
with .catch(). (rejection id: 1)
(node:14928) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.

```

After redis start:

```
{ ReplyError: NOSCRIPT No matching script. Please use EVAL.
[...]

PR #316, opened 2020-11-10
fix: handle script reinstantiation during Redis failover

rtasson:
This PR helps handle script re-instantiation during Redis failover events (in AWS at least.) When swapping instances, the script cache disappears, even while keeping the contents of the Redis cache intact. At present, `bee-queue` only loads scripts during startup, which means that after failover events this `evalsha` call fails for already running bee-queue instances. This PR attempts to catch this error, re-instantiate the scripts on the new Redis server, and try again.

An alternative and less obtuse implementation could be to utilize [`SCRIPT EXISTS`](https://redis.io/commands/script-exists) to check before every `EVALSHA`, but that doubles the round trips to Redis on every run; instead we catch this odd edgecase and only try to re-instantiate them when we encounter it.

My javascript isn't great; I'd appreciate any and all checks for correctness, code quality & style, and things I just hadn't considered. As a result, I'm filing this as a draft to receive early feedback on implementation. I'll be testing this out more fully on localhost later; at present, `npm test` passes all 140 tests.

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -830,16 +830,40 @@ class Queue extends Emitter {
       args[i] = arguments[i];
     }
 
-    return this._commandable().then((client) => {
-      // Get the sha for the script after we're ready to avoid a race condition
-      // with the lua script loader.
-      args[0] = lua.shas[name];
+    return this._commandable()
+      .then((client) => {
+        // Get the sha for the script after we're ready to avoid a race
+        // condition with the lua script loader.
+        args[0] = lua.shas[name];
 
-      const promise = helpers.deferred();
-      args.push(promise.defer());
-      client.evalsha.apply(client, args);
-      return promise;
-    });
+        const promise = helpers.deferred();
+        args.push(promise.defer());
+        client.evalsha.apply(client, args);
+        return promise;
+      })
+      .catch((err) => {
+        // In cases where the Redis server fails over, the script cache is
+        // cleared (even when the cache itself is still intact.) We detect this
+        // condition and reload the scripts.
+        if (
+          err.name === 'ReplyError' &&
+          err.message === 'NOSCRIPT No matching script. Please use EVAL.'
+        ) {
+          return lua.buildCache(this.client).then((client) => {
+            // Try again.
+            // Get the sha for the script after we're ready to avoid a race
+            // condition with the lua script loader.
+            args[0] = lua.shas[name];
+
+            const promise = helpers.deferred();
+            args.push(promise.defer());
+            client.evalsha.apply(client, args);
+            return promise;
+          });
+        } else {
+          throw err;
+        }
+      });
   }
 }
 
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -11,6 +11,8 @@ const backoff = require('./backoff');
 const EagerTimer = require('./eager-timer');
 const finally_ = require('p-finally');
 
+const scriptCachePromises = new WeakMap();
+
 class Queue extends Emitter {
   constructor(name, settings) {
     super();
@@ -114,7 +116,7 @@ class Queue extends Emitter {
     ])
       .then(() => {
         if (this.settings.ensureScripts) {
-          return lua.buildCache(this.client);
+          return (scriptCachePromises = lua.buildCache(this.client));
         }
       })
       .then(() => {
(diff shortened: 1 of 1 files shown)


rtasson:
I believe this may address https://github.com/bee-queue/bee-queue/issues/147.

lib/queue.js
```diff
@@ -830,16 +830,40 @@ class Queue extends Emitter {
...
+        // condition and reload the scripts.
+        if (
+          err.name === 'ReplyError' &&
+          err.message === 'NOSCRIPT No matching script. Please use EVAL.'
+        ) {
+          return lua.buildCache(this.client).then((client) => {
+            // Try again.
+            // Get the sha for the script after we're ready to avoid a race
+            // condition with the lua script loader.
+            args[0] = lua.shas[name];
```
shils, line 856: `args` will already have been modified in the preceding `then`, we just need to replace the last element with a fresh `helpers.deferred().defer()`

lib/queue.js
```diff
@@ -830,16 +830,40 @@ class Queue extends Emitter {
...
+
+            const promise = helpers.deferred();
+            args.push(promise.defer());
+            client.evalsha.apply(client, args);
+            return promise;
+          });
+        } else {
+          throw err;
```
shils, line 864: We should get this out of the way first, reduce indentation of the more critical code.

lib/queue.js
```diff
@@ -830,16 +830,40 @@ class Queue extends Emitter {
...
+        client.evalsha.apply(client, args);
+        return promise;
+      })
+      .catch((err) => {
+        // In cases where the Redis server fails over, the script cache is
+        // cleared (even when the cache itself is still intact.) We detect this
+        // condition and reload the scripts.
+        if (
+          err.name === 'ReplyError' &&
+          err.message === 'NOSCRIPT No matching script. Please use EVAL.'
```
skeggse, line 850: this isn't well-documented, but `node_redis` errors have a `code` field:
```suggestion
          err.code === 'NOSCRIPT'
```

```
ReplyError: NOSCRIPT No matching script. Please use EVAL.
    at parseError (/private/tmp/redis/node_modules/redis-parser/lib/parser.js:193:12)
    at parseType (/private/tmp/redis/node_modules/redis-parser/lib/parser.js:303:14) {
  command: 'EVALSHA',
  args: [ 'asdkjhdfg', 0 ],
  code: 'NOSCRIPT'
}```

File: lib/queue.js [hunk 4]
```diff
@@ -830,16 +830,40 @@ class Queue extends Emitter {
...
+        ) {
+          return lua.buildCache(this.client).then((client) => {
```
skeggse, line 852: I'm somewhat concerned that this will hammer the redis server if the client tries to execute a bunch of scripts all at once - they'll all come back with `NOSCRIPT` errors and will all separately trigger this code path. IMO we should handle that by introducing new code (either in `buildCache` or between here and `buildCache`) to memoize the promise. It'd want to be per-client, I suspect - maybe using a `WeakMap` (just storing it on the `Queue` would work too)?

skeggse (changes requested):
We might want/need to revisit [`lib/lua/index.js`](https://github.com/bee-queue/bee-queue/blob/master/lib/lua/index.js), seeing as it was written with the assumption that none of the script loading was performance-critical. I suspect it's more important now.

Also worth noting:

>An alternative and less obtuse implementation could be to utilize [`SCRIPT EXISTS`](https://redis.io/commands/script-exists) to check before every `EVALSHA`

^ this is actually how the proposed changes operate, because `buildCache` invokes `loadScriptIfMissing`, which runs `SCRIPT EXISTS`.

lib/queue.js
rtasson: this is excellent; I was nervous about including the entire string, because that could easily change in the future.

lib/queue.js
rtasson: I believe I've hit that note in the latest commit; lemme know if that's wrong.

lib/queue.js
```diff
@@ -830,16 +830,32 @@ class Queue extends Emitter {
...
+        return _evalScriptInner(client);
+      })
+      .catch((err) => {
+        // In cases where the Redis server fails over, the script cache is
+        // cleared (even when the cache itself is still intact.) We detect this
+        // condition and reload the scripts.
+        if (err.code !== 'NOSCRIPT') {
+          throw err;
+        }
+        return lua.buildCache(this.client).then((client) => {
```
skeggse, line 854: The promise produced by `buildCache` will not resolve to the client. We'll want to use `_commandable` again anyway in case the `Queue` is in the process of exiting.

rtasson: d'oh!

lib/queue.js
rtasson: [I believe `WeakMap` requires node versions >= 12 (10 behind an interpreter flag)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap#Browser_compatibility); does this change this recommendation at all? I see that you are planning deprecation of node 4 & 6 - wondering about 8/10 in this context & given that your friends at Mixmax (probably) aren't the only users of this package. I know I personally use 10 in some places still :L (tho this could be the kick in the butt to prioritize some other things.)

skeggse: I'm seeing Node `0.12`, not `12` - am I looking at something different?

lib/lua/index.js
```diff
@@ -49,12 +51,14 @@ function loadScriptIfMissing(client, scriptKey) {
...
+  if (scriptsLoaded) return scriptsLoadedPromise;
+  scriptsLoaded = true;
   // We could theoretically pipeline this, but it's pretty insignificant.
-  return readScripts().then(() =>
+  return (scriptsLoadedPromise = readScripts().then(() =>
     Promise.all(
       Object.keys(shas).map((key) => loadScriptIfMissing(client, key))
     )
-  );
+  ));
```
rtasson, line 61: @skeggse is this a reasonable way to accomplish this, in lieu of a WeakMap? I see this being done above when reading the scripts in. perhaps this isn't reasonable because it's not done at the `Client` level?

skeggse: Yeah, we'll need to do it per-Queue or per-Client - otherwise we won't correctly load scripts into auxiliary clients.

lib/queue.js
```diff
@@ -32,6 +32,8 @@ class Queue extends Emitter {
     this.bclient = null;
     this.eclient = null;
 
+    this._scriptCachePromise = new WeakMap();
```
skeggse, line 35: I suspect you're already aware but I don't see where this is used.

lib/queue.js
rtasson: nah, you're correct - I mean 0.12 and not 12. d'oh, I'm just mixed up.

rtasson:
In light of a potential rewrite, given that Mixmax has successfully accomplished the overarching task that prompted this, and a lack of time on my part to fixup this PR, I'm closing this. Thanks for the time spent reviewing! It has been good for me :heart_decoration:

PR #112, opened 2018-04-04
2.0 dev (WIP)

LewisJEllis:
I have moved everything over to ioredis, torn out all the callback stuff, and gotten all the tests back to passing! It took a while, but it's pretty much the main hurdle/technical risk for making the 2.0 plan work. That's enough to warrant a WIP branch. I will ping people here when it's less WIP and more ready for review.

Some notes/thoughts:
- Can't wait to refactor a bunch of promise stuff to async/await now that tests are back to green
- I really really like ava, and wow the tests run so fast! 3-4 seconds on my laptop.
- ioredis
  - Their auto-reconnection stuff is great - actually simplified some tests and makes me feel good about trying to test and trust connection re-establishment more
    - keeping an eye on ioredis issue 610 to cover bases/maybe repro potential issues
  - Still lots of little todos to utilize some of their nice features, cluster capabilities, etc
    - also a cool `defineCommand` capability for lua scripts - could obviate a lot of the load-script-cache stuff we've implemented
  - filed issue 614 on ioredis to resolve a confusion on status/ready detection/timing
- tearing out callbacks
  - Deleted so many instances of `if (cb) helpers.asCallback(promise, cb);`
  - new rule that your `.process` handler should just be an `async` func is great
- Moved `promise-callbacks` down to a devDependency
  - using several of them in tests so kept whole thing as devDependency
  - implementation code is basically only using `waitOn` and `withTimeout` though, so plucked those two into helpers file

Other todos:
- maybe add debug logs, ioredis's debug logs were v helpful at times
- figure out where a percent of test coverage went and get it back
- investigate viability of sharing event subscriber connections, worker connections
- bunch of readme/example updates to expunge callbacks
- redo benchmarks
- probably include some other features in 2.0 for completeness, like pluggable job data serializers and stallLimit

Refs #104

Issue #104, opened 2018-03-19
bee-queue 2.0 plan


LewisJEllis:
While investigating #103 I re-read the `process()` implementation and how we conjoin promises + callbacks for job handlers and it got me thinking a little bit - I want to get away from the callback API, and it's also just sort of that time where we get fully onboard the promises + async/await train. 

Plan:
- Use, treat, and support async/await as a first-class thing
- Stop supporting/using callbacks for anything
- No longer require `promise-callbacks` dependency
- Support only Node 8.x+ for native async/await

Pros:
- Remove promise-callbacks dependency (and save a little bit of corresponding overhead)
- Simplify some dual-pronged code paths and error handling scenarios
- Improve overall codebase readability
- Better utilize native Promise and async/await performance improvements (which have been significant in the last year)

Cons:
- Will be backwards-incompatible, with no more callback support
  - Should not require code changes for users already using promise-based handler functions
  - Users requiring the callback API can use 1.x
- Will not support old Node versions
  - Node 4 is EOL, Node 6 is old, everyone who will upgrade bee-queue to 2.0 will be on Node 8+

In general, asynchronous control flow stuff is easy to get wrong and having one API instead of two reduces the surface area for potential problems (like the confusion in #103). I also think the codebase could get a lot more readable with some `async await` mixed in and I refuse to introduce a build step.

Are there any other things that should be thought about for 2.0, since it'll be a major version? We can add other potentially-minor-version things to the milestone too, but let's especially make sure to think about changes we might want that could only go into major versions.

/cc @skeggse @bradvogel

bradvogel:
Yes - is great!

v4l3r10:
Waiting for new release!  :smile:

LewisJEllis:
Started poking into this today and realized that migrating to `ioredis` is also going to be part of this update; when #21 was originally opened, it wasn't as clear whether ioredis was a better choice, but now it's clear that ioredis has pulled ahead, and as a bonus their API supports promises out of the box.

hyeomans:
Hey `ioredis` is great, another thing that supports is [redis cluster](https://github.com/luin/ioredis#cluster) which npm `redis` does not support out of the box.

bradvogel:
+1 on ioredis. Excited for this update Lewis!

denigmus:
Hope with ioredis you will learn bee-queue, i.e. redis.createClient support automatic re-connections on redis connection lost.

hughsw:
[...]

PR #317, opened 2020-11-13
2.0: drop node 4/6, callbacks; use ioredis

### LINKED ISSUE #16: Replace node_redis with ioredis

skeggse:
I took #112 and rebased it against the default branch - phew!

I probably preserved way more of the individual commits than I needed to :grin: (edit: still reachable from 6693a0c7816b8152368bde3ef0a1188095d834a1)

Blocking TODOs:

- [ ] update README
- [ ] fix code coverage
- [x] provide some usability guidance if attempting to pass in a node_redis client
- [x] validate ioredis client abort and quit behaviors
- [ ] figure out whether we need to break anything to address #189 - likely will want to change defaults

Might release this as a 2.0 alpha so our friends over @mixmaxhq can evaluate it in wild, if they're willing.

@LewisJEllis had some fantastic ideas about next steps in #112, and I look forward to pursuing them after we've landed the simplest possible 2.0 that prevents further breakage.

I'll post a review ~in a minute~ eventually that calls out some concerns.

See also #104. Fixes #16 and closes #21.

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -71,12 +71,14 @@ class Queue extends Emitter {
       this._delayedTimer.on('trigger', () => this._activateDelayed());
     }
 
-    const makeClient = async (clientName, createNew) => {
-      const client = await redis.createClient(this.settings.redis, createNew);
+    const makeClient = async (createNew, getClientID) => {
+      const client = await redis.createClient(this.settings.redis, {
+        createNew,
+        getClientID,
+      });
       // This event gets cleaned up and removed in Queue#close for the
       // primary client if quitCommandClient is disabled.
       client.on('error', this._emitError);
-      this[clientName] = client;
       return client;
     };
 
@@ -84,7 +86,8 @@ class Queue extends Emitter {
       if (!this.settings.getEvents && !this.settings.activateDelayedJobs) {
         return;
       }
-      const eclient = await makeClient('eclient', true);
+      const eclient = await makeClient(true, false);
+      this.eclient = eclient;
       eclient.on('message', (channel, message) =>
         this._onMessage(channel, message)
       );
@@ -103,8 +106,10 @@ class Queue extends Emitter {
     // Wait for needed client connections, event subs, and Lua script caching
     this._ready = (async () => {
       await Promise.all([
-        makeClient('client', false),
-        this.settings.isWorker ? makeClient('bclient', true) : null,
+        makeClient(false, false).then((client) => (this.client = client)),
+        this.settings.isWorker
+          ? makeClient(true, true).then((bclient) => (this.bclient = bclient))
+          : null,
         makeEventClient(),
       ]);
       if (this.settings.ensureScripts) {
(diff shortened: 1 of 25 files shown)


shils:
Very exciting, I'd love to evaluate this in the @mixmaxhq wilderness.

lib/queue.js
```diff
@@ -749,9 +698,17 @@ class Queue extends Emitter {
...
     const promise = this._doStalledJobCheck();
-    if (cb) helpers.asCallback(promise, cb);
+    // TODO: what DO
+    // TODO: reintroduce tests
+    if (cb) {
+      promise.then(
+        (value) => process.nextTick(() => cb(null, value)),
+        (err) => process.nextTick(() => cb(err))
+      );
+    }
```
skeggse, line 708: So this is sad, and I would like to find another solution. The `checkStalledJobs` method supports a callback so that the library user may receive callbacks for _all_ attempted stall checks - not just the first. Removing the callback doesn't permit us to propagate any useful information, as the process won't really finish until `Queue#close` is called.

Perhaps we use events and emit an event? How do we handle errors in that case?

lib/queue.js
```diff
@@ -798,22 +755,38 @@ class Queue extends Emitter {
...
+    for (const [defer, [error, jobId]] of helpers.zip(defers, results)) {
+      if (error) {
+        defer.reject(error);
+      } else {
+        defer.resolve(jobId);
       }
-      return helpers.callAsync((done) => batch.exec(done)).then(() => errors);
-    });
+    }
+    await final;
```
skeggse, line 788: `ioredis` has abysmal support for pipelining as the pipeline commands don't return promises but instead return the pipeline for chaining. This means we have to manually identify which of the jobs each command corresponds to :\

lib/redis.js
```diff
@@ -1,79 +1,135 @@
...
-    // node_redis mutates the options object we provide it, so we clone the
-    // settings first.
-    if (typeof settings === 'object') {
-      settings = Object.assign({}, settings);
+      return client;
+    } else {
+      isNewClient = false;
     }
+  } else {
+    client = Redis.createClient(settings);
```
skeggse, line 45: This is deprecated, but we override it in tests and can't replace the bare function as easily as we can the `createClient` export.

lib/redis.js
```diff
@@ -1,79 +1,135 @@
...
+      // Fallthrough.
+    }
   }
+  disconnect(client);
+}
+
+function disconnect(client) {
+  // HACK: flushQueue is, according to ioredis, private. This is unfortunate,
+  // because we need this in order to properly clean up during brpoplpush
+  // commands.
```
skeggse, line 97: Rather gross. Other options?

lib/redis.js
```diff
@@ -1,79 +1,135 @@
...
+  return err.name === 'AbortError' || err.message === 'Connection is closed.';
+}
+
+const rBoundary = /\b/g;
+function errorCode(err) {
+  rBoundary.lastIndex = 1;
+  const { message } = err,
+    match = rBoundary.exec(err.message);
+  return err.message.slice(0, match ? match.index : undefined);
 }
```
skeggse, line 114: Probably worth submitting upstream, as this is a nice feature of NodeRedis and impacts #316

lib/redis.js
```diff
@@ -1,79 +1,135 @@
...
 
-function isClient(object) {
-  if (!object || typeof object !== 'object') return false;
+function getClientType(object) {
+  if (!object || typeof object !== 'object') return null;
   const name = object.constructor.name;
-  return name === 'Redis' || name === 'RedisClient';
+  if (name === 'Redis') return 'ioredis';
+  if (name === 'RedisClient') return 'NodeRedis';
```
skeggse, line 120: enums how do they work

lib/queue.js
```diff
@@ -663,8 +612,8 @@ class Queue extends Emitter {
 
       // invariant: in this code path, this.running < this.concurrency, always
       // after spoolup, this.running + this.queued === this.concurrency
```
skeggse, line 614: Very much would like to replace this with a more maintainable pattern that specifically handles this concurrency management.

compwright:
Has conflicts and is abandoned, closing.

PR #346, opened 2020-12-21
add repeat job  and documentation

WuHanMuMu:


skeggse (commented):
This looks like it's trying to solve the same problem as #169. I'd appreciate a brief discussion about how this approach differs from that one. In particular, I'm concerned that this implementation doesn't handle node failures well.

PR #349, opened 2020-12-30
chore: fix code samples

bradvogel:
Fixes error `TypeError: Class constructor Queue cannot be invoked without 'new'` in code samples. Also adds required setup instruction for one sample.


skeggse (approved):

skeggse:
Thanks!

PR #458, opened 2022-01-10
workaround

Gianluca-Casagrande-Stiga:
redis that is imported doesn't have a index.d.ts file, so the type import doesn't work.
I suggest this workaround in order to resolve the TypeScript compiling problem

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,8 +1,6 @@
 /// <reference types="node" />
-/// <reference types="redis" />
 
 import {EventEmitter} from 'events';
-import {ClientOpts, RedisClient} from 'redis';
 
 declare class BeeQueue<T = any> extends EventEmitter {
   name: string;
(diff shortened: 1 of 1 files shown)


bradvogel:
What's the TypeScript compilation error you're running into?

Gianluca-Casagrande-Stiga:
Hi, this is the error I get after "npx tsc" command:

> node_modules/bee-queue/index.d.ts:2:23 - error TS2688: Cannot find type definition file for 'redis'.
> 
> 2 /// <reference types="redis" />
>                         ~~~~~
> 
> node_modules/bee-queue/index.d.ts:5:39 - error TS7016: Could not find a declaration file for module 'redis'. '/..path_to_my_project.../node_modules/bee-queue/node_modules/redis/index.js' implicitly has an 'any' type.
>   Try `npm i --save-dev @types/redis` if it exists or add a new declaration (.d.ts) file containing `declare module 'redis';`
> 
> 5 import {ClientOpts, RedisClient} from 'redis';

I think this is because the redis version used by bee-queue 1.4.0 doesn't have TS definition

PR #461, opened 2022-01-18
Support cancellable jobs

ramiel:
A new event is exposed: `removed` that emits when a job gets removed. This way clients can listen to it and execute some action to cancel their jobs

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -25,6 +25,7 @@ declare class BeeQueue<T = any> extends EventEmitter {
   on(ev: 'job retrying', fn: (jobId: string, err: Error) => void): this;
   on(ev: 'job failed', fn: (jobId: string, err: Error) => void): this;
   on(ev: 'job progress', fn: (jobId: string, progress: any) => void): this;
+  on(ev: 'job removed', fn: (jobId: string) => void): this;
 
   ready(): Promise<this>;
   ready(cb?: (err: Error | null) => void): Promise<this>;
@@ -113,6 +114,7 @@ declare namespace BeeQueue {
     on(ev: 'retrying', fn: (err: Error) => void): this;
     on(ev: 'failed', fn: (err: Error) => void): this;
     on(ev: 'progress', fn: (progress: any) => void): this;
+    on(ev: 'removed', fn: () => void): this;
 
     setId(id: string): this;
     retries(n: number): this;
--- a/test/job-test.js
+++ b/test/job-test.js
@@ -161,6 +161,19 @@ describe('Job', (it) => {
       t.is(await hget(queue.toKey('jobs'), job.id), null);
     });
 
+    it('should warn about job being removed', async (t) => {
+      const {makeJob} = t.context;
+
+      const job = await makeJob();
+      return new Promise((resolve) => {
+        job.on('removed', () => {
+          t.true(true);
+          resolve();
+        });
+        job.remove();
+      });
+    });
+
     it('should work with a callback', async (t) => {
       const {queue, makeJob} = t.context;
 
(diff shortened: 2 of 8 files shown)


bradvogel:
I don't believe this will work as intended. `job.remove()` and `queue.removeJob` aren't meant to be called when the job is currently being processed. What if I job is being processed by another worker?

ramiel:
Yes I read the documentation:

> This may have unintended side-effect, e.g. if the job is currently being processed by another worker, so only use this method when you know it's safe.

The idea about this event is to handle side effects in the worker so that, if the job is removed when it's being processed, something can be done.

At the moment nothing prevents the developer to call this method when the worker is processing the job (just a warning in the documentation) and still this can happen in some unintended way. This event let the developer handle this case.

Anyway, if this goes against the way you want bee to work it's fine, just close this PR. just want to point out that not having cancellable jobs is a big missing feature in bee that maybe should be solved (regardless of this PR)

test/job-test.js
```diff
@@ -161,6 +161,19 @@ describe('Job', (it) => {
       t.is(await hget(queue.toKey('jobs'), job.id), null);
     });
 
+    it('should warn about job being removed', async (t) => {
```
bradvogel, line 164: How does the test assert the warning?

index.d.ts
```diff
@@ -113,6 +114,7 @@ declare namespace BeeQueue {
     on(ev: 'retrying', fn: (err: Error) => void): this;
     on(ev: 'failed', fn: (err: Error) => void): this;
     on(ev: 'progress', fn: (progress: any) => void): this;
+    on(ev: 'removed', fn: () => void): this;
```
bradvogel, line 117: Needs to be documented in the readme. Add a code example showing how jobs should handle their 'removed' event internally. Note that it requires the 'getEvents' option.

bradvogel (changes requested):

test/job-test.js
ramiel: I meant "notified", like an event is fired when the job gets removed. 

So, it's not clear to me: I though you didn't want to accept the PR, from your previous comment

index.d.ts
ramiel: I added some documentation in the readme

bradvogel:
@ramiel Thanks for the continued work on this. I tested it locally and was unable to get it to work. Perhaps you could help me.

Here's what I doing:
1. Check out branch https://github.com/ramiel/bee-queue/pull/1
2. Run `cd examples/pingpong`
3. Run `node pong.js`
4. In another shell, run `node ping.js`

Expected result: I see "REMOVED" get logged in the output for pong.js.
Actual result: I don't see it logged.

ramiel:
Sorry for being so late on this. I noticed the same and I think it's because `job` is a different instance, so the event emitter cannot work. The same event on the queue is working instead

```js
// Not working
  job.on('removed', () => {
    console.log('REMOVED');
  });

// Working
  pingQueue.on('job removed', (jobId) => {
    console.log('REMOVED', jobId);
  });
```
You can check the code here: https://github.com/ramiel/bee-queue/blob/158ff916f78c48ef47ea1a21c478694add5cd10f/examples/pingpong/pong.js

Maybe you can help on this?

ramiel:
For example, if you put, in pong file

```js
job.on('progress', (i) => {
    console.log('PROGRESS', i);
  });
```

this is not working either. I think this is because of bee architecture: the job is not the same instance, so the event emitter is different.

avin-kavish:
How about inserting some value to a redis structure that is being listened on by all workers.

compwright:
Not working and appears abandoned. If you want to continue working on this, please rebase your branch and resubmit a new PR that addresses all the issues above.

PR #192, opened 2020-03-24
fix unpacking too many values, taken from existing PR-#123

Tob0t:
Since there is no progress on #123 I created another PR.

FYI: The commit from @mattbornski was taken.

Since our queues stopped working if there are more than 8k jobs with the same timestamp, we needed a fix asap. I know that the PR was not merged since there could be a further improvements. However as @mattbornski mentioned, better this fix with blocking redis for short time than processing no jobs at all.

We are already using this code on our production system since we needed a fix asap.

Disclaimer: I'm not a lua programmer, so feel free to adapt the PR ;)

PR #550, opened 2022-11-22
fix: queues stop working when there are more than 8k jobs with the same timestamp

compwright:
Resubmitting https://github.com/bee-queue/bee-queue/pull/192 from @Tob0t and @mattbornski to rebase and re-run CI checks:

> Since there is no progress on #123 I created another PR.
> 
> FYI: The commit from @mattbornski was taken.
> 
> Since our queues stopped working if there are more than 8k jobs with the same timestamp, we needed a fix asap. I know that the PR was not merged since there could be a further improvements. However as @mattbornski mentioned, better this fix with blocking redis for short time than processing no jobs at all.
> 
> We are already using this code on our production system since we needed a fix asap.
> 
> Disclaimer: I'm not a lua programmer, so feel free to adapt the PR ;)

### DIFF
--- a/lib/lua/raiseDelayedJobs.lua
+++ b/lib/lua/raiseDelayedJobs.lua
@@ -10,13 +10,21 @@ returns number of jobs raised and the timestamp of the next job (within the near
 local now = tonumber(ARGV[1])
 
 -- raise any delayed jobs that are now valid by moving from delayed to waiting
-local raising = redis.call("zrangebyscore", KEYS[1], 0, ARGV[1])
-local numRaising = #raising
+local offset = 0
+-- There is a default 1024 element restriction
+local count = 1000
+while true do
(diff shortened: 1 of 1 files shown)


bradvogel (approved):
Approving.

It looks like this was [approved here](https://github.com/bee-queue/bee-queue/pull/123#pullrequestreview-136017280) but with [this concern](https://github.com/bee-queue/bee-queue/pull/123#pullrequestreview-349581747). However, I agree with the comment in the PR description that it's better to block Redis than to not process at all.

compwright:
@tmeasday care to add a review before we move on this?

tmeasday:
Seems reasonable but it would be good to have a test I guess (I'm no expert on how these tests work anymore, or ever really).

bradvogel:
@tmeasday You're absolutely right. I complete overlooked the fact that there wasn't a test 🤦‍♂️. I think this is OK to skip one for now since it's unlikely the original author is interested in adding one now.

PR #650, opened 2023-06-03
fix(examples): Parse request params as number

myogh:
Initially, they were parsed as strings, and thus the result was '23' instead of 5.


compwright (approved):

compwright:
Thank you for your contribution! I had to change the commit message to fix failing tests. Resubmitting as #651.

PR #708, opened 2023-10-26
allow manual redis connection

billyen2012:
Hey, this is a super awesome queue library.

I'm working on a `next.js` project and found that whenever I try to build the project, the build process will failed because `Queue` is try to connect to the redis host for some reasons.

Not sure if it is because the logic of connection to redis host is in the constructor of the `Queue`,  but be able manually connect to redis seems to solve the problem.

So here is the pull request basically allowing user to manual connect to redis by calling `queue.connect()` after `Queue` instance is created. 
- test is added and README is updated accordingly;
- there is also minor code refactor.

### DIFF
--- a/README.md
+++ b/README.md
@@ -322,6 +322,7 @@ const queue = new Queue('test', {
   removeOnSuccess: false,
   removeOnFailure: false,
   redisScanCount: 100,
+  autoConnect: true,
 });
 ```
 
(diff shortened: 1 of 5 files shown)


compwright:
@billyen2012 also please note the failing CI checks. You need to edit the commit message for all commits on this branch to comply with this standard: https://github.com/conventional-changelog/commitlint/#what-is-commitlint

billyen2012:
> Thank you for this contribution! I think it would be a bit more user-friendly to call the new option `autoConnect` and default it to `true`. If you don't want auto connection then you would manually set the option to `false` and call `queue.connect()`. Would you make this change for me?

Suggested change has been applied

billyen2012:
> @billyen2012 also please note the failing CI checks. You need to edit the commit message for all commits on this branch to comply with this standard: https://github.com/conventional-changelog/commitlint/#what-is-commitlint

Opps, I was unaware that I suppose to use `npm run ci` for all the test.

should be okay now
image


I also notice that sometimes the test will failed because when all the test suite are executed together, it will just make redis dropping connection. Not sure if it was just the issue of my redis, but I will suggest the following change to make the test more stable. (this is not in this PR)


image

compwright (approved):

PR #773, opened 2024-02-16
feat: added a typecheck to prevent further errors

dogabudak:
I was using bee-queue and I just realised a stupid mistake I made after long time of research. A more descriptive error would save me a lot of time.

JobId can only be number or string and in a node.js project it is really easy to pass `null` or even an object as an argument. Maybe this can save some other peoples time. 
If you pass any other argument other then string/number you get a generic error from redis-client which is not super helpful.

Screenshot 2024-02-16 at 22 16 22

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -19,13 +19,17 @@ class Job extends Emitter {
 
   static fromId(queue, jobId, cb) {
     const promise = queue
-      ._commandable()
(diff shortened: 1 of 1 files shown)


compwright (changes requested):
Thank you for this! It looks like you have some formatting issues in the code change, could you fix those for me?

Issue #481, opened 2022-04-13
Upgrade node redis from v2.* to v4.*


KiranI-Bsw:
I am using node redis v4.0.2 for my project. However, bee-queue uses v2.* and I am unable to reuse my redis client from v4. Gives somethig like this. I am not dropping out of v4 bcoz I can use promises with node redis itself. Due to this issue I have to open a new connection for bee-queue queue whereas I should have been able to use one. Any suggestion on how to deal with this?

```
node_modules/bee-queue/node_modules/redis/lib/utils.js:65
            var snake_case = elem.replace(/[A-Z][^A-Z]/g, '_$&').toLowerCase();
                                  ^

RangeError: Maximum call stack size exceeded
    at String.replace (<anonymous>)
    at clone (/Users/bswmac013/Documents/GitHub/redis-matchmaking/matchmaker-consumer/node_modules/bee-queue/node_modules/redis/lib/utils.js:65:35)
    at clone (/Users/bswmac013/Documents/GitHub/redis-matchmaking/matchmaker-consumer/node_modules/bee-queue/node_modules/redis/lib/utils.js:71:32)
    at clone (/Users/bswmac013/Documents/GitHub/redis-matchmaking/matchmaker-consumer/node_modules/bee-queue/node_modules/redis/lib/utils.js:71:32)
    at clone (/Users/bswmac013/Documents/GitHub/redis-matchmaking/matchmaker-consumer/node_modules/bee-queue/node_modules/redis/lib/utils.js:71:32)
    at clone (/Users/bswmac013/Documents/GitHub/redis-matchmaking/matchmaker-consumer/node_modules/bee-queue/node_modules/redis/lib/utils.js:71:32)
    at clone (/Users/bswmac013/Documents/GitHub/redis-matchmaking/matchmaker-consumer/node_modules/bee-queue/node_modules/redis/lib/utils.js:71:32)
    at clone (/Users/bswmac013/Documents/GitHub/redis-matchmaking/matchmaker-consumer/node_modules/bee-queue/node_modules/redis/lib/utils.js:71:32)
    at clone (/Users/bswmac013/Documents/GitHub/redis-matchmaking/matchmaker-consumer/node_modules/bee-queue/node_modules/redis/lib/utils.js:71:32)
    at clone (/Users/bswmac013/Documents/GitHub/redis-matchmaking/matchmaker-consumer/node_modules/bee-queue/node_modules/redis/lib/utils.js:71:32)

```

[...]

PR #886, opened 2025-03-14
add notes in docs to inform people latest version of node_redis is not supported

1mike12:
I don't want others to waste time debugging something like myself and so many other people have done, because they're not made aware of a pretty large caveat. 

i know there's a discussion and issue created but this info imo should live next to where most people will be tripped up by it, not burried in issues. 

https://github.com/bee-queue/bee-queue/issues/481


compwright:
Can't get this to format right, closing.

PR #1030, opened 2026-07-20
fix: correct off-by-one in getJobs default end index

tsushanth:
## Problem

`Queue#getJobs` defaults to `{ start: 0, end: 100 }` for list/sorted-set
queue types (`waiting`, `active`, `delayed`). Both Redis `LRANGE` and
`ZRANGE` use **inclusive** bounds, so this call:

```
lrange key 0 100
```

returns **101** elements (indices 0 through 100 inclusive), not 100.

This is inconsistent with the `size: 100` default used for the set-based
queue types (`failed`, `succeeded`), where the intent is clearly to return
at most 100 jobs by default.

## Fix

Change the default `end` from `100` to `99` so that a no-argument
`getJobs('waiting')` call returns at most 100 jobs, matching the behaviour
of `getJobs('failed')` and the documented intent.

Callers who pass an explicit `page.end` are unaffected.

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -387,7 +387,7 @@ class Queue extends Emitter {
    * @param {Number=} page.start Start of query range for waiting/active/delayed
    *   queue types. Defaults to 0.
    * @param {Number=} page.end End of query range for waiting/active/delayed
-   *   queue types. Defaults to 100.
+   *   queue types. Defaults to 99.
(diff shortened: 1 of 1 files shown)


acid4you:
I have maked  a package that name is queue-jobs-worker it's support multiple adapters see my main profile @rafidahmed870

